PR #5, opened 2020-05-13
Mod1 Week2 Independent Challenge_Ruthie Rabinovitch

rrabinovitch:


lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    return true if @price > 500000
+    false
+  end
```
memcmahon, line 17: This is great - you could also rely on the return value of the condition:

```ruby
def above_market_average?
    price > 500000
end
```

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def rooms_from_category(category)
+    @rooms.find_all do |room|
```
memcmahon, line 20: great use of `find_all`

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def area
+    house_area = 0
+
+    @rooms.map do |room|
+      house_area += room.area
+    end
+
+    house_area
+  end
```
memcmahon, line 33: with this pattern, `each` would be better than `map`.  You could also look into the `sum` method

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def rooms_by_category
+    rooms_by_category = {}
+
+    @rooms.map do |room|
+      rooms_by_category[room.category] = rooms_from_category(room.category)
+    end
+
+    rooms_by_category
+  end
```
memcmahon, line 58: Great job with this method - also look into the method `group_by`

test/house_test.rb
```diff
@@ -0,0 +1,157 @@
...
+  end
+
+  def test_it_can_identify_if_above_market_average
+    house_1 = House.new("$400000", "123 sugar lane")
+    house_2 = House.new("$500000", "456 pepper street")
+    house_3 = House.new("$500001", "789 salt avenue")
+
+    assert_equal false, house_1.above_market_average?
+    assert_equal false, house_2.above_market_average?
+    assert_equal true, house_3.above_market_average?
```
memcmahon, line 48: great job testing multiple scenarios!

rrabinovitch:
thanks for the feedback @memcmahon !

PR #6, opened 2020-05-13
Gaby Méndez

gabichuelas:
Woo hoo! All iterations complete & all tests pass. 
I feel like I could use more practice / guidance writing better commit messages.

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,5 +1,5 @@
 class House
-  attr_reader :price, :address, :rooms
+  attr_reader :price, :address, :rooms, :rooms_by_category
   def initialize(price, address)
     @price = price[1..-1].to_i
     @address = address
@@ -15,11 +15,13 @@ class House
   end
 
   def rooms_from_category(category)
-    room_category_hash =
+    rooms_by_category[category]
+  end
+
+  def rooms_by_category
     @rooms.group_by do |room|
       room.category
     end
-    room_category_hash[category]
   end
 
   def area
@@ -32,4 +34,13 @@ class House
     Hash.new.merge("price" => @price, "address" => @address)
   end
 
+  def price_per_square_foot
+    @price.fdiv(area).round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by do |room|
+      room.area
+    end
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,12 @@
+class House
+  attr_reader :price, :address, :rooms
+  def initialize(price, address)
+    @price = price[1..-1].to_i
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end 
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,5 +1,5 @@
 class House
-  attr_reader :price, :address, :rooms, :rooms_by_category
+  attr_reader :price, :address, :rooms
   def initialize(price, address)
     @price = price[1..-1].to_i
     @address = address
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -8,5 +8,28 @@ class House
 
   def add_room(room)
     @rooms << room
-  end 
+  end
+
+  def above_market_average?
+    true if @price > 500000
+  end
+
+  def rooms_from_category(category)
+    room_category_hash =
+    @rooms.group_by do |room|
+      room.category
+    end
+    room_category_hash[category]
+  end
+
+  def area
+    @rooms.reduce(0) do |sum, room|
+      sum += room.area
+    end
+  end
+
+  def details
+    Hash.new.merge("price" => @price, "address" => @address)
+  end
+
 end
--- a/test/house_test.rb
+++ b/test/house_test.rb
@@ -6,6 +6,10 @@ require './lib/house'
 class HouseTest < Minitest::Test
   def setup
     @house = House.new("$400000", "123 sugar lane")
+    @room_1 = Room.new(:bedroom, 10, '13')
+    @room_2 = Room.new(:bedroom, 11, '15')
+    @room_3 = Room.new(:living_room, 25, '15')
+    @room_4 = Room.new(:basement, 30, '41')
   end
 
   def test_it_exists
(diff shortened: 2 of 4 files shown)


timomitchel:
Nice job committing throughout the challenge.

lib/house.rb
```diff
@@ -0,0 +1,46 @@
+class House
+  attr_reader :price, :address, :rooms
+  def initialize(price, address)
+    @price = price[1..-1].to_i
```
timomitchel, line 4: Nice job here, but this isn't as flexible as it could be. Imagine that we have a price passed in without a `$` or with two `$$` then this will cause problems. `price.delete("$").to_i` is the most flexible implementation.

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    true if @price > 500000
```
timomitchel, line 14: No need for the guard statement here. `@price > 500000` is all we need.

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def rooms_from_category(category)
+    rooms_by_category[category]
+  end
+
+  def rooms_by_category
```
timomitchel, line 21: Nice job using the right enumerable for the job here!

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+    @rooms.group_by do |room|
+      room.category
+    end
+  end
+
+  def area
+    @rooms.reduce(0) do |sum, room|
```
timomitchel, line 28: Nice job using reduce correctly here, but this is the perfect use case for the `.sum` enum: 
```
@rooms.sum do |room|
room.area
end
```
which reads a little bit more clearly and is less work.

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+      sum += room.area
+    end
+  end
+
+  def details
+    Hash.new.merge("price" => @price, "address" => @address)
```
timomitchel, line 34: This certainly works, and is pretty clean, but the following is a bit more readable and avoids calling `#merge`:
`{"price" => @price, "address" => @address}`

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def price_per_square_foot
+    @price.fdiv(area).round(2)
```
timomitchel, line 38: Nice! `#fdiv` isn't something I've seen before! This is cool!

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by do |room|
```
timomitchel, line 42: Ah, so close. We needed to throw a `#reverse` on this return value in order to get this array in descending order, which is what the interaction pattern called for. But, `sort_by` is the correct enum for the job:
```
@rooms.sort_by do |room|
room.area
end.reverse
```

test/house_test.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def test_it_has_total_area
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+
+    assert_equal 1900, @house.area
+  end
```
timomitchel, line 61: Really solid testing to this point. Great use of a `#setup` method.

test/house_test.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def test_it_has_details
+    assert_instance_of Hash, @house.details
```
timomitchel, line 64: I like the way you improvised the testing here because minitest was probably giving you problems when you attempted to pass the hash to the `#assert_equal` method. You can accomplish this by using a variable OR parentheses:
```
def test_house_details
    house = House.new("$400000", "123 sugar lane")
    expected = {"price" => 400000, "address" => "123 sugar lane"}
    assert_equal expected, house.details
  end
```

test/house_test.rb
```diff
@@ -0,0 +1,95 @@
...
+    assert_equal 210.53, @house.price_per_square_foot
+  end
+
+  def test_can_sort_rooms_by_area
+    @house.add_room(@room_4)
+    @house.add_room(@room_1)
+    @house.add_room(@room_3)
+    @house.add_room(@room_2)
+
+    assert_equal [@room_1, @room_2, @room_3, @room_4], @house.rooms_sorted_by_area
```
timomitchel, line 83: The interaction pattern calls for these in reverse order. Attention to detail here. If you matched up the object ids with the interaction pattern, then you could deduce this.

test/house_test.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def test_can_group_rooms_by_category
+    @house.add_room(@room_4)
+    @house.add_room(@room_1)
+    @house.add_room(@room_3)
+    @house.add_room(@room_2)
+
+    assert_instance_of Hash, @house.rooms_by_category
```
timomitchel, line 92: Again, I like the workaround testing for this hash return.

test/room_test.rb
```diff
@@ -3,22 +3,33 @@
...
   def test_it_can_get_area
-    skip
+    # skip
     room1 = Room.new(:bedroom, 10, '13')
     room2 = Room.new(:living_room, 15, '12')
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_it_is_painted
```
timomitchel, line 29: Be careful with assert and refute. This gave you a pssing test on line 30, even though a nil is returned.

lib/house.rb
gabichuelas: During the challenge review I was def like "wait, why did I overcomplicate this?" there must have been a weird train of thought happening that I can't remember. ha!

lib/house.rb
gabichuelas: Oh sweet, I didn't realize it would return false without doing if/else statements.

lib/house.rb
gabichuelas: Didn't know #sum existed until after the challenge review!!

lib/house.rb
gabichuelas: I totally didn't remember learning that, but found it in my own notes while looking for something else, haha!

lib/house.rb
gabichuelas: Yeah, I remember wondering what order y'all wanted us to sort_by, but for some reason I didn't see the list of object ID's as something I should reference for the order, instead I saw it as "this should return an array of room objects". Whamp whamp

test/house_test.rb
gabichuelas: @timomitchel 
I actually never tried to pass the hash itself as the "expected" result, just seemed more efficient to check for the object class of @house.details. 

Would you say checking for the hash itself is preferable? 

My thought was that it wouldn't scale bc the hash might be something crazy long! and I know also that the way the computer likes to return it is different than the way I would write it.. right? (like attribute order and whatnot)

test/house_test.rb
gabichuelas: @timomitchel 
Wait but isn't the object ID different every time the object gets created? I don't remember what the interaction pattern looked like now, but would I have been able to tell what order the rooms were in from the ID number?

test/room_test.rb
gabichuelas: Thanks - learned this at some point from Megan a couple days ago and definitely changed a bunch of my Battleship tests because of it!!

test/house_test.rb
timomitchel: My bad on the delay. If you can check for exact value of the entire hash then you should (if it's not crazy long). Order of your key/value pairs will not matter. As long as they are all in the hash, the test will pass.

test/house_test.rb
timomitchel: Yes, it is different, but the interaction pattern is built in a way that you can see which objects are held by which variable, and then the return values are in that specific order.

PR #7, opened 2020-05-13
Jessye Ejdelman

ejdelsztejn:
* Created Room class
* Created House test
* Created House class

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -5,4 +5,8 @@ class House
     @address = address
     @rooms = []
   end
+
+  def add_room(room)
+    @rooms << room
+  end
 end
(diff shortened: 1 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def rooms_by_category
+    room_types = Hash.new(0)
+    @rooms.each do |room|
+      room_types[room.category] = []
+    end
+    @rooms.each do |room|
+      room_types[room.category] << room
+    end
+    room_types
```
KathleenScriver, line 49: Great job figuring out a solution to this problem. One thing to note though, if you are iterating over the same array twice, can you combine those two actions so you only have to iterate over the rooms array once? Iterating over an array is pretty expensive in terms of processing power, so the more we can reduce iteration, the better.

PR #8, opened 2020-05-13
Grant Dempsey real_estate_2005

GDemps:
Cheers!

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -39,12 +39,8 @@ class House
   end
 
   def rooms_sorted_by_area
-    @rooms.sort_by { |room| room.area }
-  end
-
-  def rooms_by_category
-    @rooms
-
+    sorted = @rooms.sort_by { |room| room.area }
+    sorted.reverse
   end
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -43,4 +43,11 @@ class House
     sorted.reverse
   end
 
+  def rooms_by_category
+    bedroom = rooms_from_category(:bedroom)
+    living_room = rooms_from_category(:living_room)
+    basement = rooms_from_category(:basement)
+    { bedroom: bedroom, living_room: living_room, basement: basement}
+  end
+
 end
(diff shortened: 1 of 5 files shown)


lib/house.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def rooms_from_category(category)
+    @rooms.find_all do |room|
+      room.category == category
+    end
+  end
+
+  def area
+    @rooms.sum do |room|
+      room.area.to_f
```
KathleenScriver, line 29: The interaction pattern actually did not indicate that area should be a float.

lib/house.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  def details
+    { "price" => @price, "address" => @address}
```
KathleenScriver, line 34: Great job here.

lib/house.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def price_per_square_foot
+    (price/area).round(2)
```
KathleenScriver, line 38: oh, I see. I think perhaps you had #area return a float to help you here in this method. You could have just done area.to_f though here to force this answer to return a float.

lib/house.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def rooms_sorted_by_area
+    sorted = @rooms.sort_by { |room| room.area }
```
KathleenScriver, line 42: Great use of the sort_by method.

lib/house.rb
```diff
@@ -0,0 +1,53 @@
...
+    sorted.reverse
+  end
+
+  def rooms_by_category
+    bedroom = rooms_from_category(:bedroom)
+    living_room = rooms_from_category(:living_room)
+    basement = rooms_from_category(:basement)
+    { bedroom: bedroom, living_room: living_room, basement: basement}
```
KathleenScriver, line 50: Great job building this out and using helper methods. I do want you to keep in mind what Megan was saying in the review session that using this particular helper method results in a lot of array iteration which is an expensive process. How else could you accomplish this method using less iteration?

PR #9, opened 2020-05-13
Kiera Allen

KieraAllen:
1. Completed Iteration 1 - Iteration 3
2. Did not finish `rooms_by_category` method of Iteration 4

* misspelled 'spelling' as 'speling' in first commit message (EMBARRASSING)

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -8,4 +8,8 @@ class House
     @address = address
     @rooms = []
   end
+
+  def add_room(room)
+    @rooms << room
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,7 @@
+class House
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+  end
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,7 +1,9 @@
 class House
+  attr_reader :price,
+              :address
 
   def initialize(price, address)
-    @price = price
+    @price = price.gsub("$", '').to_i
     @address = address
   end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -16,4 +16,10 @@ class House
   def above_market_average?
     price >= 500000
   end
+
+  def rooms_from_category(type_of_room)
+    rooms.find_all do |room|
+      room.category == type_of_room
+    end
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -12,4 +12,8 @@ class House
   def add_room(room)
     @rooms << room
   end
+
+  def above_market_average?
+    price >= 500000
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -22,4 +22,12 @@ class House
       room.category == type_of_room
     end
   end
+
+  def area
+    house_area = 0
+    rooms.each do |room|
+       house_area += room.area
+    end
+    house_area
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -37,4 +37,8 @@ class House
     details["address"] = address
     details
   end
+
+  def price_per_square_foot
+    (@price.to_f / self.area).round(2)
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -45,4 +45,13 @@ class House
   def rooms_sorted_by_area
     rooms.sort_by { |room| room.area }.reverse
   end
+
+  def rooms_by_category
+    sort_them_rooms = {}
+    sort_them_rooms[:bedroom] =
+    sort_them_rooms[:living_room] =
+    sort_them_rooms[:basement] = 
+    rooms.sort_by { |room| room.category }
+    sort_them_rooms
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -41,4 +41,8 @@ class House
   def price_per_square_foot
     (@price.to_f / self.area).round(2)
   end
+
+  def rooms_sorted_by_area
+    rooms.sort_by { |room| room.area }.reverse
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,9 +1,11 @@
 class House
   attr_reader :price,
-              :address
+              :address,
+              :rooms
 
   def initialize(price, address)
     @price = price.gsub("$", '').to_i
     @address = address
+    @rooms = []
   end
 end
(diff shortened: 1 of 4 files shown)


KathleenScriver:
Great job with committing so often. I would encourage you to do different messages though. Maybe something like `Adds total house area method` after you've done the tests.

test/room_test.rb
```diff
@@ -9,16 +9,35 @@ def test_it_exists
...
     room2 = Room.new(:living_room, 15, '12')
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_it_has_length_and_width
+    room = Room.new(:bedroom, 10, '13')
+
+    assert_equal 10, room.length
+    assert_equal 13, room.width
```
KathleenScriver, line 27: Be careful here. The interaction pattern DID want a room to be initialize with a length and width, but we did NOT ask for methods of length and width. Basically, we don't want to expose that information if it's not asked for.

test/room_test.rb
```diff
@@ -9,16 +9,35 @@ def test_it_exists
...
+  end
+
+  def test_it_starts_not_painted
```
KathleenScriver, line 30: Great test name, very specific and precise.

test/house_test.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  def test_it_can_read_price_and_address_attributes
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal 400000, house.price
+    assert_equal "123 sugar lane", house.address
+  end
+
+  def test_it_starts_with_no_rooms
```
KathleenScriver, line 20: Great test name.

test/house_test.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    assert_equal [room_1, room_2], house.rooms
+  end
+
+  def test_it_is_not_above_market_average_if_price_is_under_500000
+    house = House.new("$400000", "123 sugar lane")
+    house2 = House.new("$600000", "456 sugar lane")
+
+    assert_equal false, house.above_market_average?
+    assert_equal true, house2.above_market_average?
```
KathleenScriver, line 43: Great job testing for both the `true` and `false` conditions of this method.

test/house_test.rb
```diff
@@ -0,0 +1,131 @@
...
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal 1900, house.area
+  end
+
+  def test_it_has_details
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal ({"price" => 400000, "address" => "123 sugar lane"}), house.details
```
KathleenScriver, line 82: Great job testing the hash here.

test/house_test.rb
```diff
@@ -0,0 +1,131 @@
...
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
+    assert_equal [room_4], house.rooms_from_category(:basement)
```
KathleenScriver, line 60: Great job here just testing two cases. If it works for 2 different cases, we can be confident this method should work for almost any case.

lib/room.rb
```diff
@@ -1,12 +1,24 @@
-class Rom
+class Room
+  attr_reader :category,
+              :length,
```
KathleenScriver, line 3: Here again, be careful. The interaction pattern did not ask for methods of length and width so we don't want to expose those methods.

lib/house.rb
```diff
@@ -0,0 +1,57 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize(price, address)
+    @price = price.gsub("$", '').to_i
```
KathleenScriver, line 7: Great job figuring out a string method that will accomplish this.

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    price >= 500000
```
KathleenScriver, line 17: Careful here. The interaction pattern says if the house is exactly 500000, this method should return false.

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def rooms_from_category(type_of_room)
+    rooms.find_all do |room|
```
KathleenScriver, line 21: Great use of find_all enumerable.

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+      room.category == type_of_room
+    end
+  end
+
+  def area
+    house_area = 0
+    rooms.each do |room|
+       house_area += room.area
+    end
+    house_area
```
KathleenScriver, line 31: Great, this works! Play with this method using other enumerables to see if you can be more efficient with your code.

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def details
+    details = {}
+    details["price"] = price
+    details["address"] = address
+    details
```
KathleenScriver, line 38: Great job adding key/value pairs to this hash. Can you think of another, more literal way to accomplish this?

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def price_per_square_foot
+    (@price.to_f / self.area).round(2)
```
KathleenScriver, line 42: You shouldn't need to call `self` in front of area here. Great job changing one of the integers to a float to ensure you get decimal places.

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def rooms_sorted_by_area
+    rooms.sort_by { |room| room.area }.reverse
```
KathleenScriver, line 46: Great job here figuring out the order this should be sorted and using the correct enumerable.

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def rooms_by_category
+    sort_them_rooms = {}
+    sort_them_rooms[:bedroom] =
+    sort_them_rooms[:living_room] =
+    sort_them_rooms[:basement] = 
+    rooms.sort_by { |room| room.category }
+    sort_them_rooms
```
KathleenScriver, line 55: Keep working with this one! I bet with a little more research and thinking about how hashes work, you can get this one.

PR #10, opened 2020-05-13
Alex Desjardins

moosehandlr:


timomitchel:
Nice job committing throughout!

lib/house.rb
```diff
@@ -0,0 +1,54 @@
+class House
+
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize (price, address)
+    @price = price
+    @address = address
+    @rooms = []
```
timomitchel, line 10: Nice, this is the setup we were looking for.

lib/house.rb
```diff
@@ -0,0 +1,54 @@
+class House
+
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize (price, address)
```
timomitchel, line 7: There should not be a space between `initialize` and the parentheses

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+    @price = price
+    @address = address
+    @rooms = []
+
```
timomitchel, line 11: No need for this empty line.

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def price
+    @price.gsub(/\$/,'').to_i
```
timomitchel, line 15: Nice use of `gsub` here.

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+              :rooms
+
+  def initialize (price, address)
+    @price = price
+    @address = address
+    @rooms = []
+
+  end
+
+  def price
```
timomitchel, line 14: We don't ever want an `attr_reader` and a method named the same thing, because the `attr_reader` is never accessed. It is overridden by the method definition.

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+    @price.gsub(/\$/,'').to_i
+  end
+
+  def add_room(new_room)
```
timomitchel, line 18: Nice job here. This is a common pattern you will see in independent challenges.

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+    @rooms << new_room
+  end
+
+  def above_market_average?
+    if price > 500000
```
timomitchel, line 23: Great job setting up the right condition here, but the `if` is redundant. We only need this method to have `price > 500000` which will always return a true or false.

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(room_type)
+    @room_type = room_type
```
timomitchel, line 31: No need to define this instance variable here. We have access to the `room_type` parameter inside the block. Also, `select` and `find_all` are synonymous, so this is the correct enumerable for this method because we are looking for a subset of the original array. I tend to use `find_all` as I think it reads better than select in terms of what it's doing.

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+    @rooms.select do |room|
+      room.category == @room_type
+    end
+  end
+
+  # def house_area
+  #   @length * @width.to_i
+  # end
+
+  def house_area
```
timomitchel, line 41: This is on the right tack, but not quite there. One thing to note is we always want to name our methods exactly the way they are named in the interaction pattern, because if we don't then it will fail the tests that we instructors use to grade these. This was a great use case for the `.sum` enumerable which adds all values that are returned for each iteration of the block. This will give you what you want if you change the code to:
```
 @rooms.sum do |room|
      room.room_area
    end
```
We don't want to name our block variable `|room_area|` because that isn't what it represents. It represents a `Room` object, so we call it `|room|`

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+    @rooms.each do |room_areas|
+      room_areas.room.room_areas.join
+    end
+  end
+
+  def details = {}
```
timomitchel, line 47: Not familiar with this syntax. Definitely not a ruby convention.

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+    price_details = Hash.new(price)
```
timomitchel, line 48: There will be some use cases ahead for initializing a hash with a default value like so `Hash.new(0)`, but this isn't one of them. This method should read: 
```
{"price" => @price, "address" => @address}
```
based on the interaction pattern.

lib/room.rb
```diff
@@ -1,12 +1,31 @@
-class Rom
+class Room
+
+  attr_reader :category,
+              :length,
+              :width,
+              :room_area
```
timomitchel, line 6: No need for the `:room_area` attr or instance variable.

lib/room.rb
```diff
@@ -1,12 +1,31 @@
...
 
   def initialize(category, length, width)
     @category = category
-    @length = 10
+    @length = length
     @width = width
+    @room_area = length * width.to_i
```
timomitchel, line 12: This functionality for `@room_area` should live in a method called `#area`

lib/room.rb
```diff
@@ -1,12 +1,31 @@
...
+
+  def paint
+    true
   end
+
+  # def is_painted?
+  #   if paint
+  #     false
+  #   end
+  # end
```
timomitchel, line 28: Yea, this was a tough piece of the interaction pattern, but essentially the big takeaway is recognizing both command and query methods used in conjunction with an instance variable. Let me know if you have more questions on this.

test/house_test.rb
```diff
@@ -0,0 +1,82 @@
...
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+
+    house = House.new("$400000", "123 sugar lane")
+
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+
+    assert_equal [room_1, room_2], house.rooms
```
timomitchel, line 37: Really solid testing up to this point.

test/house_test.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def test_if_house_is_above_market_average
```
timomitchel, line 40: Nice job here. I would also add a test that results in a `true` value returned to make this testing more robust.

test/house_test.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    house = House.new("$400000", "123 sugar lane")
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
+    assert_equal [room_4], house.rooms_from_category(:basement)
```
timomitchel, line 60: Great job here!

test/house_test.rb
```diff
@@ -0,0 +1,82 @@
...
+    house = House.new("$400000", "123 sugar lane")
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+
+
+    assert_equal 1900, house.house_area
```
timomitchel, line 78: Also a great test here.

test/room_test.rb
```diff
@@ -9,16 +9,20 @@ def test_it_exists
...
     assert_equal :bedroom, room.category
   end
 
   def test_it_can_get_area
-    skip
     room1 = Room.new(:bedroom, 10, '13')
     room2 = Room.new(:living_room, 15, '12')
-    assert_equal 130, room1.area
-    assert_equal 180, room2.area
+    assert_equal 130, room1.room_area
```
timomitchel, line 19: It's never a good idea to change up the tests that are already written for you.

test/room_test.rb
```diff
@@ -9,16 +9,20 @@ def test_it_exists
...
+    assert_equal 180, room2.room_area
   end
+
+  # def test_if_room_is_painted
```
timomitchel, line 23: We definitely want a test for this interaction. I don't recommend moving on from the first iteration until you've got it working properly.

timomitchel:
Nice job committing throughout!

PR #11, opened 2020-05-13
Saryn Mooney Real Estate 2005BE

sarynm12:


lib/house.rb
```diff
@@ -0,0 +1,47 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = '$400000'
```
memcmahon, line 5: Not every house will cost $400000!  It is better to use the argument that you passed in!  `@price = price`

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+    rooms << room
+  end
+
+  def above_market_average?
+    if price.to_i > $500000.to_i
+      true
+    else
+      false
+    end
+  end
```
memcmahon, line 20: Good job with this method!

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def rooms_from_category(category)
+    rooms.select do |room|
+      room.category == :bedroom
+    end
+  end
```
memcmahon, line 26: We won't always be looking for bedrooms - better to use the argument you sent in:

```ruby
rooms.select do |room|
  room.category == category
end
```

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+    room3 = Room.new(:living_room, 25, '15')
+    room4 = Room.new(:basement, 30, '41')
+    house = House.new("$400000", "123 sugar lane")
+    house.add_room(room1)
+    house.add_room(room2)
+    house.add_room(room3)
+    house.add_room(room4)
+    165 + 130 + 375 + 1230
+
+    #I know I messed this up. I did the math on paper, but was not getting the same value when I added it in Ruby! Sorry!
```
memcmahon, line 40: Its ok, though I do want to make sure you understand that when we are looking for information based on a collection, it is always best to iterate over that collection - these won't always be the rooms in all houses.

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def details
+    @price
+    @address
+  end
```
memcmahon, line 46: On the right track with the info we are looking for, but we want it in a hash:

```ruby
{price: @price, address: @address}
```

test/house_test.rb
```diff
@@ -0,0 +1,75 @@
...
+require './lib/room'
+require 'pry'
+
+class HouseTest < Minitest::Test
+
+  def test_it_exists
+    house = House.new("$400000", "123 sugar lane")
+    assert_instance_of House, house
+  end
+
```
memcmahon, line 13: You are missing tests for `.address` and `.price`

test/house_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    room2 = Room.new(:bedroom, 11, '15')
+    room3 = Room.new(:living_room, 25, '15')
+    room4 = Room.new(:basement, 30, '41')
+    house = House.new("$400000", "123 sugar lane")
+    house.add_room(room1)
+    house.add_room(room2)
+    house.add_room(room3)
+    house.add_room(room4)
+
+    assert_equal [room1, room2], house.rooms_from_category(:bedroom)
```
memcmahon, line 45: This test passes only because you are looking for bedrooms - because your method is hardcoded for bedrooms, if you look for any other category of rooms, the test will fail and show you where you went wrong.

test/house_test.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def test_it_returns_area
```
memcmahon, line 48: This test looks good, but the method itself needs work 😉

test/house_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    room1 = Room.new(:bedroom, 10, '13')
+    room2 = Room.new(:bedroom, 11, '15')
+    room3 = Room.new(:living_room, 25, '15')
+    room4 = Room.new(:basement, 30, '41')
+    house = House.new("$400000", "123 sugar lane")
+    house.add_room(room1)
+    house.add_room(room2)
+    house.add_room(room3)
+    house.add_room(room4)
+    assert_equal house, house.details
```
memcmahon, line 73: The assertion we are looking for here is:

```ruby
assert_equal ({price: "$400000", address: "123 sugar lane"}), @house.details
```

test/room_test.rb
```diff
@@ -1,24 +1,88 @@
...
+    assert_equal 120, room2.area
+  end
+
+  def test_it_knows_if_painted
+    room = Room.new(:bedroom, 10, '13')
+
+    assert_equal false, room.is_painted?
   end
+
+
```
memcmahon, line 32: you are missing a test for the `.paint` method

PR #12, opened 2020-05-13
Nick Edwin

NickEdwin:
I tried! :(

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,31 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize(price, address)
+    @price = price.gsub(/\D/,'').to_i
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    rooms << room
+  end
+
+  def above_market_average?
+    if price > 500000
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(type)
+    return_list == []
+    rooms.map do |style|
+      if type == style
+        
+      end
+    end
+end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -1,12 +1,25 @@
-class Rom
+class Room
+  attr_reader :category,
+              :length,
+              :width,
+              :painted
 
   def initialize(category, length, width)
     @category = category
-    @length = 10
+    @length = length
     @width = width
+    @painted = false
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_painted?
+    @painted
+  end
+
+  def paint
+    @painted = !@painted
   end
 end
--- a/test/house_test.rb
+++ b/test/house_test.rb
@@ -1,0 +1,56 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/room'
+require './lib/house'
+
+class HouseTest < Minitest::Test
+  def setup
+    house = House.new("$400000", "123 sugar lane")
+  end
+
+  def test_it_exists
+    house = House.new("$400000", "123 sugar lane")
+    assert_instance_of House, house
+  end
+
+  def test_it_has_atrributes
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal 400000, house.price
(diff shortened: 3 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,31 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize(price, address)
+    @price = price.gsub(/\D/,'').to_i
```
memcmahon, line 7: Great job with this price method!

lib/house.rb
```diff
@@ -0,0 +1,31 @@
...
+    rooms << room
+  end
+
+  def above_market_average?
+    if price > 500000
+      true
+    else
+      false
+    end
+  end
```
memcmahon, line 22: This totally works, and you could also rely on the return value of the condition:

```ruby =
def above_market_average?
  price > 500000
end
```

lib/house.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def rooms_from_category(type)
+    return_list == []
+    rooms.map do |style|
+      if type == style
+        
+      end
+    end
```
memcmahon, line 30: Good start on this method.  One thing that would help is to call your block variable `room`, so that it is more clear what type of object you have access to.

And, on line 25, you are using the comparison == where you should be using the assignment =.

Also, when we are looking for a subset of a collection, we want to use `each` or `find_all` instead of `map`

```ruby
def rooms_from_category(type)
  return_list = []
  rooms.each do |room|
    if room.category == type
      return_list << room
    end
  end
  return_list
end
```

lib/room.rb
```diff
@@ -1,12 +1,25 @@
-class Rom
+class Room
+  attr_reader :category,
+              :length,
+              :width,
+              :painted
```
memcmahon, line 5: Since we have the method `is_painted?` below, we don't need an attr_reader for the `@painted` instance variable.

lib/room.rb
```diff
@@ -1,12 +1,25 @@
...
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_painted?
+    @painted
+  end
+
+  def paint
+    @painted = !@painted
```
memcmahon, line 23: Great!

test/house_test.rb
```diff
@@ -0,0 +1,56 @@
...
+    house = House.new("$400000", "123 sugar lane")
+    assert_instance_of House, house
+  end
+
+  def test_it_has_atrributes
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal 400000, house.price
+    assert_equal "123 sugar lane", house.address
+  end
+
```
memcmahon, line 21: Don't forget to test `.price`!

test/house_test.rb
```diff
@@ -0,0 +1,56 @@
...
+  def test_it_has_no_rooms
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal true, house.rooms.empty?
+  end
+
+  def test_it_can_add_rooms
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    house.rooms
```
memcmahon, line 31: You don't need this line - but otherwise this test for adding rooms looks great!

test/house_test.rb
```diff
@@ -0,0 +1,56 @@
...
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+    house.rooms_from_category(:bedroom)
+    assert_equal [room_1, room_2], house.rooms_from_category
```
memcmahon, line 53: Anytime you call this method, you need to send it an argument!

```ruby
assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
```

PR #13, opened 2020-05-13
Kwibe Merci

jKwibe:


test/room_test.rb
```diff
@@ -9,16 +9,30 @@ def test_it_exists
...
+
+    assert_equal false , room.is_painted?
+  end
+
+  def test_it_paints
+    room = Room.new(:bedroom, 10, '13')
+
+    assert_equal false, room.is_painted?
+    room.paint
+    assert_equal true , room.is_painted?
```
KathleenScriver, line 36: Great job testing the initial state of is_painted?, then calling another method, and then testing the resulting state of is_painted? That's a great pattern.

test/room_test.rb
```diff
@@ -9,16 +9,30 @@ def test_it_exists
...
   def test_it_can_get_area
-    skip
+    # skip
     room1 = Room.new(:bedroom, 10, '13')
     room2 = Room.new(:living_room, 15, '12')
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_it_is_painted
```
KathleenScriver, line 25: This test name is a little misleading since you're ensuring the method returns `false`. I would love to have seen this called something like `test_it_starts_not_painted` or something similar.

test/house_test.rb
```diff
@@ -0,0 +1,163 @@
...
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_instance_of House, house
+  end
+
+  def test_it_has_address_and_price
+    # skip
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal "$400000", house.price
```
KathleenScriver, line 19: Careful with this one. The interaction pattern asks that the return value of `#price` is an integer, not a string.

test/house_test.rb
```diff
@@ -0,0 +1,163 @@
...
+    assert_equal "123 sugar lane", house.address
+  end
+
+  def test_it_has_rooms
```
KathleenScriver, line 23: You are actually testing that it starts with no rooms, so would love to see this test named something that reflects that more.

test/house_test.rb
```diff
@@ -0,0 +1,163 @@
...
+
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+
+    assert_equal 2 , house.rooms.size
+    assert_instance_of Room, house.rooms[0]
+    assert_instance_of Room, house.rooms[1]
```
KathleenScriver, line 42: Can also do an assertion testing the actual array, something like: 
```
assert_equal [room_1, room_2], house.rooms
```
This would ensure that the 2 rooms that you just added are indeed the rooms in the house's rooms array.

test/house_test.rb
```diff
@@ -0,0 +1,163 @@
...
+  end
+
+  def test_it_is_above_average
+    # skip
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal false, house.above_market_average?
```
KathleenScriver, line 49: When we have a method that returns a boolean value, we want to be sure to test that the method can return both `true` and `false`.

test/house_test.rb
```diff
@@ -0,0 +1,163 @@
...
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal 4, house.rooms.size
+
+    assert_equal 2, house.rooms_from_category(:bedroom).size
+    assert_equal 1, house.rooms_from_category(:basement).size
```
KathleenScriver, line 69: This doesn't actually test that the _correct_ house is the one in the array. I would rather see a more specific assertion to make sure the room instance with that category was the room in the array. For example: 
```
assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
```

test/house_test.rb
```diff
@@ -0,0 +1,163 @@
...
+    assert_equal 4, house.rooms.size
+
+    assert_equal 1900, house.area
+  end
+
+  def test_it_has_details
+    # skip
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal ["price", "address"], house.details.keys
```
KathleenScriver, line 95: This does test the keys of the details array, but it doesn't test that the values are correct. You'll want to investigate a bit more in how to test a hash with minitest. You have to either store the hash in a variable, or wrap it in parenthesis in the assertion.

test/house_test.rb
```diff
@@ -0,0 +1,163 @@
...
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+
+    house.add_room(room_4)
+    house.add_room(room_1)
+    house.add_room(room_3)
+    house.add_room(room_2)
+
+    assert_equal :bedroom, house.rooms_sorted_by_area.last.category
+    assert_equal :basement, house.rooms_sorted_by_area.first.category
```
KathleenScriver, line 128: The interaction pattern is expecting this method to return a hash. It doesn't look like you're testing for that.

test/house_test.rb
```diff
@@ -0,0 +1,163 @@
...
+  end
+end
+
+
+# pry(main)> house = House.new("$400000", "123 sugar lane")
```
KathleenScriver, line 133: Don't forget to remove all these comments before submitting final work.

lib/room.rb
```diff
@@ -1,12 +1,21 @@
-class Rom
-
+class Room
+  attr_reader :category
```
KathleenScriver, line 2: Great job paying attention to the fact that we didn't ask for methods of `length` and `width`.

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+    @rooms = []
+    @details = ["price", "address"]
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    @price.gsub("$", "").to_i > 500000
```
KathleenScriver, line 16: Good job researching how to get this one done.

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def rooms_from_category(room_category)
+    @rooms.select{|room| room.category == room_category}
```
KathleenScriver, line 20: Great job here and great use of the correct enumerable.

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def area
+    @rooms.map{|room| room.area}.inject(){|sum, area| sum + area}
```
KathleenScriver, line 24: A few things here:
1. Great job using different enumerables, it's clear you are understanding how these work. 
2. Can you chain enumerables and blocks? Yes. Should you? I err on the side of no since it makes the code much less readable and harder to debug and refactor later.
3. Do you need to use inject here? If you have an array of integers, which `map` here would return, we have a great array method that would sum those up for us, so we don't need to use all the processing power of inject.

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def details
+   Hash[@details.zip([@price, @address])]
```
KathleenScriver, line 28: Looks like you might want to do some more exploration on how to build hashes.

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def price_per_square_foot
+    (@price.gsub("$", "").to_i / self.area.to_f).round(2)
```
KathleenScriver, line 32: This is a great example of repeated code. You used this same gsub method above on the price. If you're doing that multiple times, there might be a better place to convert the price to an integer so you only have to run the gsub method once. 
Great job using the to_f and round methods!

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by{|area| self.area}
```
KathleenScriver, line 36: So close! Pay attention to your block variables though. What is each element of the @rooms array?

lib/house.rb
jKwibe: Thanks! I think I was kinda panicking because of the time so I brought in the inject enumerable. I'll try to look at the code later and try to minimize the complexity of the code used.

lib/house.rb
jKwibe: Yap! a little bit of digging in.

lib/house.rb
jKwibe: I think I should create a method that converts the string on the fly. I do not know how to ask this in writing, personally I think that converting the string on the initialize method would be bad because we might need the string later on in our code. What do you think about that?

PR #14, opened 2020-05-13
Judith Pillado 

judithpillado:


test/room_test.rb
```diff
@@ -3,22 +3,29 @@
...
-    assert_equal 130, room1.area
-    assert_equal 180, room2.area
+    assert_equal 130, @room1.area
+    assert_equal 180, @room2.area
   end
+
+  def test_if_it_is_painted
+    assert_equal false, @room1.is_painted?
+    @room1.paint
+    assert_equal true, @room1.is_painted?
```
KathleenScriver, line 28: Love how you tested this one. You tested an initial query, then ran a command, then ran a second query to ensure that the affects of your command method had the desired results.

test/house_test.rb
```diff
@@ -0,0 +1,88 @@
...
+  def test_it_exists
+    assert_instance_of House, @house
+  end
+
+  def test_it_has_readable_attributes
+    assert_equal 400000, @house.price
+    assert_equal "123 sugar lane", @house.address
+  end
+
+  def test_it_starts_with_no_rooms
```
KathleenScriver, line 26: Great naming of your test here. Very descriptive of the initial state.

test/house_test.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def test_it_can_add_rooms
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    assert_equal [@room_1, @room_2], @house.rooms
+  end
+
+  def test_if_it_is_above_market_average
+    assert_equal false, @house.above_market_average?
+    assert_equal true, @house2.above_market_average?
```
KathleenScriver, line 38: Great job testing both of the boolean cases this method could return.

test/house_test.rb
```diff
@@ -0,0 +1,88 @@
...
+  def test_house_area
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+    assert_equal 1900, @house.area
+  end
+
+  def test_house_details
+    assert_equal ({"price" => 400000, "address" => "123 sugar lane"}), @house.house_details
```
KathleenScriver, line 60: Great job testing this hash.

lib/room.rb
```diff
@@ -1,12 +1,23 @@
-class Rom
+class Room
+  attr_reader :category
```
KathleenScriver, line 2: Great job exposing only the method the interaction pattern asked for.

lib/house.rb
```diff
@@ -0,0 +1,59 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price.tr('$', '').to_i
```
KathleenScriver, line 5: Great job figuring out a string method to convert this.

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+    @address = address
+    @rooms = []
+    @above_market_average = false
```
KathleenScriver, line 8: I don't think we need to make this an instance variable. What if the price of the house changes? Then we have another variable we would have to update to ensure it's reading the correct state. It would be better to keep this dynamic by just using your method.

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    if price > 500000
```
KathleenScriver, line 16: This method just needs to return true or false if the house's price evaluates to greater than or less than 500000. Think about the return value of the statement `price > 500000`, do we need to include an if/else block? Hint: you did this perfectly with the Room#is_painted? method.

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+      true
+    else
+      @above_market_average
+    end
+  end
+
+  def rooms_from_category(category)
+    rooms.find_all do |room|
```
KathleenScriver, line 24: Great use of the find_all enumerable.

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+      room.category == category
+    end
+  end
+
+  def area
+    sum = 0
+    rooms.each do |room|
+      sum += room.area
+    end
+    sum
```
KathleenScriver, line 34: Great job using #each to accomplish this goal. I challenge you to think about other enumerables that would accomplish this for you more efficiently.

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def house_details
```
KathleenScriver, line 37: Be careful on what you name your method. You'll want to name the method exactly what the interaction pattern is asking for. This method should be called `details`.

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+    details = Hash.new
+    details["price"] = @price
+    details["address"] = @address
+    details
```
KathleenScriver, line 41: Great job building this hash. Can you think of another, more literal way to build this hash?

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def price_per_square_foot
+    (price.to_f / area).round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by {|room| room.area}.reverse
```
KathleenScriver, line 49: Great job using this sort_by enumerable.

lib/house.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def rooms_by_category
+    type_of_room = {}
+    @rooms.each do |room|
+      room.category
```
KathleenScriver, line 55: Keep working with this one. How can you dynamically add keys and values to the hash you created on line 53?

PR #15, opened 2020-05-13
Michael Evans

mjevans113:
Independent Challenge #2

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -42,4 +42,9 @@ class House
     house_details = {"price" => @price.gsub(/\D/,'').to_i,
                      "address" => @address}
   end
+
+  def price_per_square_foot
+    ppsf = @price.gsub(/\D/,'').to_i / area.to_f
+    ppsf.round(2)
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -52,8 +52,11 @@ class House
     @rooms.sort_by { |room| -room.area }
   end
 
-  # def rooms_by_category
-  #   by_category = {}
-  #   @rooms.each do |room|
-  # end
+  def rooms_by_category
+    by_cat = {}
+    @rooms.each do |room|
+      by_cat[room.category] = rooms_from_category(room.category)
+    end
+    by_cat
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -11,4 +11,22 @@ class House
   def add_room(room)
     @rooms << room
   end
+
+  def above_market_average?
+    if @price.gsub(/\D/,'').to_i > 500000
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(room_category)
+    rooms_same_category = []
+    @rooms.each do |room|
+      if room.category == room_category
+        rooms_same_category << room
+      end
+    end
+    rooms_same_category
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -37,4 +37,9 @@ class House
     end
     house_area
   end
+
+  def details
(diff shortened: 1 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+    @rooms << room
+  end
+
+  def above_market_average?
+    if @price.gsub(/\D/,'').to_i > 500000
+      true
+    else
+      false
+    end
+  end
```
memcmahon, line 21: This totally works, and you could also rely on the return value of the condition:

```ruby =
def above_market_average?
  @price.gsub(/\D/,'') > 500000
end
```

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def rooms_from_category(room_category)
+    rooms_same_category = []
+    @rooms.each do |room|
```
memcmahon, line 25: Great use of `each` - you could also use `find_all`

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+    rooms_same_category
+  end
+
+  def area_of_house
+    house_area = 0
+    @rooms.each do |room|
+      house_area += room.area
+    end
+    house_area
+  end
```
memcmahon, line 39: This method works, but was supposed to be called `area` 😢

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def details
+    house_details = {"price" => @price.gsub(/\D/,'').to_i,
```
memcmahon, line 42: This does not _need_ to be assigned to a variable :)

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+                     "address" => @address}
+  end
+
+  def price_per_square_foot
+    ppsf = @price.gsub(/\D/,'').to_i / area_of_house.to_f
```
memcmahon, line 47: Try not to use acronyms as variable names - in 3 days if I ask you what `ppsf` means, will you remember?

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+    ppsf.round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by { |room| -room.area }
```
memcmahon, line 52: yay math 🙃

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def rooms_by_category
+    by_cat = {}
+    @rooms.each do |room|
+      by_cat[room.category] = rooms_from_category(room.category)
+    end
+    by_cat
+  end
```
memcmahon, line 61: GREAT job with this method!

test/house_test.rb
```diff
@@ -0,0 +1,126 @@
...
+  def test_it_exists
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_instance_of House, house
+  end
+
+  def test_it_has_readable_attributes
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal "$400000", house.price
```
memcmahon, line 17: Keep a close eye on the interaction pattern - we wanted `price` to return an integer

mjevans113: I did not notice that! Thank you for pointing that out

lib/house.rb
mjevans113: I know =\ I did this primarily to avoid confusing myself lol. I wanted make a clear distinction between this area method and the room area method, but it does break the expectation for the interaction pattern.

mjevans113:
Thanks for your feedback, Megan! I will continue to work on identifying when it is/isn't necessary to use an if statement to return conditional values :)

PR #16, opened 2020-05-13
Arique (she/they) #2005

Arique1104:


lib/house.rb
```diff
@@ -0,0 +1,44 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms,
+              :price
+
+  def initialize(price, address)
+    @price = price.tr('$', '').to_i
```
KathleenScriver, line 8: Great job playing with methods to see how to accomplish this.

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    if @price < 400001
```
KathleenScriver, line 18: Think about what the statement `@price < 400001` returns. Is the if statement necessary?  Also, I know the interaction pattern didn't ask for this, but a method that returns a boolean should be able to handle both cases. This method would never return true. Something to think about when writing more robust tests.

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+      false
+    end
+  end
+
+#   def rooms_from_category(category)
+#     @rooms.each do |room|
+#       room.category
+#     # new_category_search = []
+#     #     @rooms.each do |room|
+#     #         new_category_search << if room.category == category
```
KathleenScriver, line 28: You're on the right track here, just need to change up the order of how this is all happening.

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+#           # end
+#
+# require 'pry'; binding.pry
+#
+#
+#   end
+    def rooms
```
KathleenScriver, line 35: Were you attempting to make a method called `rooms` here? What is your attr_reader doing for you?

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+      @rooms.each do ||
+
+    end
+    def area
+        @house.rooms.length *
+        @house.rooms.width
```
KathleenScriver, line 41: Think about the return values here when you do method chaining. What is the return value of `@house.rooms`? And what data type is it? This should return an array and we don't have a method called `#width` that we can call on an array. Think about how you can access every element of this array.

lib/room.rb
```diff
@@ -1,12 +1,24 @@
-class Rom
```
KathleenScriver, line 1: Good job reading the errors to fix this.

lib/room.rb
```diff
@@ -1,12 +1,24 @@
...
-
+class Room
+attr_reader :category,
```
KathleenScriver, line 2: Watch your spacing here. This should be indented since it's contained inside the class.

lib/room.rb
```diff
@@ -1,12 +1,24 @@
...
     @width = width
+    @is_painted = false
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_painted?
```
KathleenScriver, line 16: Good job with naming conventions here.

test/room_test.rb
```diff
@@ -3,22 +3,33 @@
...
-    assert_equal :bedroom, room.category
+    assert_equal :bedroom, @room.category
+    assert_equal 10, @room.length
+    assert_equal '13', @room.width
   end
 
   def test_it_can_get_area
-    skip
-    room1 = Room.new(:bedroom, 10, '13')
-    room2 = Room.new(:living_room, 15, '12')
```
KathleenScriver, line 20: I know you are using a setup here, but it's good to have two instances to test to ensure your code is dynamic and can handle different values. Something to think about for future iterations.

test/room_test.rb
```diff
@@ -3,22 +3,33 @@
...
-    assert_equal 130, room1.area
-    assert_equal 180, room2.area
+    # skip
+    assert_equal 130, @room.area
   end
+
+  def test_if_room_is_painted?
+    assert_equal false, @room.is_painted?
+    @room.paint
+    assert_equal true, @room.is_painted?
```
KathleenScriver, line 31: Good job testing the initial state of `is_painted?`, applying a method, and then testing that `is_painted?` has changed. That is a great pattern to follow.

test/room_test.rb
```diff
@@ -3,22 +3,33 @@
...
-    room1 = Room.new(:bedroom, 10, '13')
-    room2 = Room.new(:living_room, 15, '12')
-    assert_equal 130, room1.area
-    assert_equal 180, room2.area
+    # skip
+    assert_equal 130, @room.area
   end
+
+  def test_if_room_is_painted?
+    assert_equal false, @room.is_painted?
```
KathleenScriver, line 29: Don't forget `assert` and refute` for minitest assertions. Those would work great here.

test/house_test.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def test_it_has_rooms
+    assert_equal [], @house.rooms
+
+    assert_equal [@room_1], @house.add_room(@room_1)
+
+    assert_equal [@room_1, @room_2], @house.add_room(@room_2)
+
+    assert_equal [@room_1, @room_2], @house.rooms
```
KathleenScriver, line 33: Great testing the initial state of #rooms, then applying a method that should change that array, then testing that method again. Great job.

test/house_test.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    assert_equal [@room_1, @room_2, @room_3], @house.add_room(@room_3)
+
+    assert_equal [@room_1, @room_2, @room_3, @room_4], @house.add_room(@room_4)
+  end
+
+  def test_it_is_above_market_average?
+    assert_equal false, @house.above_market_average?
```
KathleenScriver, line 41: Would love to see the other branch here, that this method could return `true` as well.

PR #18, opened 2020-05-13
Garrett gerdsen

ggerdsen:


timomitchel:
Nice job committing throughout the challenge!

lib/house.rb
```diff
@@ -0,0 +1,78 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
```
timomitchel, line 9: Nice setup up to this point, but no need for the `@above_market_average = false` line.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    @above_market_average = false
+  end
+
+  def rooms
```
timomitchel, line 13: This method is redundant, because this is the exact same functionality an `attr_reader` provides.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    @rooms
+  end
+
+  def add_room(room)
```
timomitchel, line 17: Nice job. This is a common pattern in challenges.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    # prior to attempting to remove first char
+    if @price.slice(0) == "$"
+      @price.slice!(0)
+    end
+    if @price.to_i > 500000
+      @above_market_average = true
+    else
+      @above_market_average
+    end
+  end
```
timomitchel, line 32: This one is a bit over-engineered. The string to integer conversion should happen in initialize, and the most effective way to do this is: `@price.delete("$").to_i`
Then, this method could simply read: `@price > 500000` which will always return the correct boolean based on our condition.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def rooms_from_category(type)
```
timomitchel, line 34: This one works, but you're using `#select` incorrectly. You're using it the way you should use an `#each`. Whenever we set up our accumulator (`new_room_array`) outside of the block, then it's likely a use case for an each. `select` and `find_all` are the exact same, and they both take a condition in the block (without any need for an if). If the condition is met for an object in the array, then that object is returned as part of the subset of the original array. If the condition is not met, then that object is disregarded and we continue iterating. So, it essentially does what you did with the `if` condition under the hood. This would look like: 
```
@rooms.select do |room|
      room.category == type
end
```

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    new_room_array = []
+    @rooms.select do |room|
+      if room.category == type
+        new_room_array << room
+      end
+    end
+    new_room_array
+  end
+
+  def area
```
timomitchel, line 44: Again, this one is a bit more complex than it needs to be, and it's a great use case for the `.sum` emumerable: 
```
@rooms.sum do |room|
room.area
end
```

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    @rooms.select do |room|
+      length_array << room.length
+      width_array << room.width
+      area_adder += length_array[counter] * width_array[counter]
+      counter += 1
+    end
+    area_adder
+  end
+
+  def details
```
timomitchel, line 58: Nice job here, but no need to set this `hash` variable as it's never used moving forward.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    hash = {"price" => @price, "address" => @address}
+  end
+
+  def price_per_square_foot
```
timomitchel, line 62: Again, this is a bit over-engineered and a lot of the work from `#area` is repeated. If the conversion for `@price` to an int had already been done, then this method could simply read: 
```
(@price.to_f / area.to_f).round(2)
```
This way you use the `#area` methods return value to avoid duplicating work.

lib/room.rb
```diff
@@ -1,12 +1,27 @@
-class Rom
+class Room
+  attr_reader :category,
+              :is_painted,
+              :area,
```
timomitchel, line 4: No need for these two `attr_reader`s

lib/room.rb
```diff
@@ -1,12 +1,27 @@
...
+              :length,
+              :width
 
   def initialize(category, length, width)
     @category = category
-    @length = 10
-    @width = width
+    @length = length.to_i
+    @width = width.to_i
+    @is_painted = false
```
timomitchel, line 12: Nice job with this setup.

lib/room.rb
```diff
@@ -1,12 +1,27 @@
...
     @length * @width
   end
+
+  def is_painted?
+    @is_painted
+  end
+
+  def paint
+    @is_painted = true
+  end
```
timomitchel, line 25: Nice job with this interaction.

test/house_test.rb
```diff
@@ -0,0 +1,87 @@
...
+    assert_equal "123 sugar lane", house.address
+  end
+
+  def test_it_has_rooms
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal [], house.rooms
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    assert house.add_room(room_1)
+    assert house.add_room(room_2)
```
timomitchel, line 31: Close, but we don't care about the return value of `#add_room`. Notice it isn't shown in the interaction pattern. We wanted to assert that the value of `#rooms` had changed:
`assert_equal [room_1, room_2], house.rooms`

test/house_test.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def test_it_has_market_average_working
+    house = House.new("$400000", "123 sugar lane")
+    refute house.above_market_average?
```
timomitchel, line 36: No need to run this twice. Also, be careful with `refute` and `assert`, because they can give false positives in your tests. Assert will pass for any value that is truthy, and refute will pass for any value that is falsy, which includes `nil`. It's better to do:
`assert_equal false, house.above_market_average?` OR
`assert_equal true, house.above_market_average?` with a different setup, say a house object with a `@price` above 500000

test/house_test.rb
```diff
@@ -0,0 +1,87 @@
...
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+    assert house.rooms_from_category(:bedroom)
```
timomitchel, line 50: This is not what we want here. It passes, but it's a poor test. 
This should read: 
`assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)`
Your test simply confirms that a truthy value is returned, which is any value other than false or nil.

test/house_test.rb
```diff
@@ -0,0 +1,87 @@
...
+    assert house.rooms_from_category(:basement)
```
timomitchel, line 65: No need to conflate the testing of this functionality with your area test.

test/house_test.rb
```diff
@@ -0,0 +1,87 @@
...
+    assert_equal 1900, house.area
```
timomitchel, line 66: This is a good assertion.

PR #19, opened 2020-05-13
Cam C - IC #2

ckccameron:


timomitchel:
Nice job committing throughout, but don't worry about branching and making pull requests during independent challenges.

lib/house.rb
```diff
@@ -0,0 +1,37 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
```
timomitchel, line 10: Nice, exactly the setup we were hoping for for house.

lib/house.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def price
+    @price.to_s.gsub(/[$]/,'').to_i
```
timomitchel, line 13: Nice use of gsub, but no need for the `.to_s` as we already have a string.

lib/house.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def add_room(room)
+    @rooms << room
```
timomitchel, line 17: Solid, and a common pattern you will see in ICs

lib/house.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def above_market_average
+    if self.price > 500000
```
timomitchel, line 21: Nice job with this method, but the `if` is redundant. `self.price > 500000` will always return us the correct boolean. Or even shorter: `@price > 500000`

lib/house.rb
```diff
@@ -0,0 +1,37 @@
...
+      false
+    end
+  end
+
+  def rooms_from_category(category)
+    @rooms[category].category
+    end
+  end
+
+  def area
```
timomitchel, line 33: This is so close! `.sum` was the best enumerable for this job:
```
@rooms.sum do |room|
      room.area
    end
```

Or you could get this to work property by calling `.sum` on the return value of the `.map`:
```
@rooms.map do |room|
      room.area
    end.sum
```
but this is essentially iterating twice, which is less performant.

lib/room.rb
```diff
@@ -1,12 +1,29 @@
-class Rom
+require 'pry'
+class Room
+  attr_reader :category
```
timomitchel, line 3: So this doesn't error, but the only `attr_reader` you have access to is `:category` because you forgot the commas. So, `:length` and `:width` are not defined properly.

lib/room.rb
```diff
@@ -1,12 +1,29 @@
...
+              :length
+              :width
 
   def initialize(category, length, width)
     @category = category
     @length = 10
```
timomitchel, line 9: We needed to change the way we set `@length` to `@length = length`

lib/room.rb
```diff
@@ -1,12 +1,29 @@
...
     @width = width
+    @is_painted = false
```
timomitchel, line 11: Nice job with this instance variable initial boolean value.

lib/room.rb
```diff
@@ -1,12 +1,29 @@
...
+  def length
+    @length
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_painted?
```
timomitchel, line 22: Great job with the `is_painted?` and `paint` methods.

test/house_test.rb
```diff
@@ -0,0 +1,84 @@
...
+    house = House.new("$400000", "123 sugar lane")
+
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+
+    assert_equal [room_1, room_2], house.rooms
+  end
```
timomitchel, line 43: Great testing on all of these tests.

test/house_test.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def test_if_price_above_market_average
+    house = House.new("$400000", "123 sugar lane")
+
```
timomitchel, line 47: Whenever we have a boolean method, it's worthwhile to test both a true and false return value.

test/house_test.rb
```diff
@@ -0,0 +1,84 @@
...
+    assert_equal false, house.above_market_average
+  end
+
+  def test_it_can_categorize_rooms
```
timomitchel, line 51: Solid test setup here!

test/house_test.rb
```diff
@@ -0,0 +1,84 @@
...
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal 1900, house.area
```
timomitchel, line 82: Your testing looks good throughout.

lib/house.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def above_market_average
+    if self.price > 500000
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(category)
```
timomitchel, line 28: Not quite. We needed to iterate on all of our rooms and find a subset that matched the category param. `find_all` is the best enumerable for this job.

PR #20, opened 2020-05-13
Lito White Real Estate Pull Request

ljwhite:
Lito White
Real Estate Project
Pull Request

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,15 @@
+class House
+
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price.tap { |str| str.delete!('$') }.to_i
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room 
+  end
+
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,6 +1,6 @@
 class House
 
-  attr_reader :price, :address, :rooms
+  attr_reader :price, :address, :rooms, :details
 
   def initialize(price, address)
     @price = price.tap { |str| str.delete!('$') }.to_i
@@ -9,7 +9,34 @@ class House
   end
 
   def add_room(room)
-    @rooms << room 
+    @rooms << room
+  end
+
+  def above_market_average?
+    false
+  end
+
+  def rooms_from_category(category)
+    arr = []
+    self.rooms.each do |room|
+      if room.category == category
+        arr << room
+      else
+      end
+    end
+    arr
+  end
+
+  def area
+    sq_footage = 0
+    self.rooms.each do |room|
+      sq_footage += room.area
+    end
+    sq_footage
+  end
+
+  def details
+    {"price" => @price, "address" => @address}
   end
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,3 +1,5 @@
+require '../lib/room'
+
 class House
 
   attr_reader :price, :address, :rooms, :details
@@ -36,7 +38,15 @@ class House
   end
 
   def details
-    {"price" => @price, "address" => @address}
+    @details = {"price" => @price, "address" => @address}
+  end
+
+  def price_per_square_foot
+    (self.price / self.area.to_f).round(2)
+  end
+
+  def rooms_sorted_by_area
+    self.rooms.sort_by {|room| room.area }
   end
 
 end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -1,12 +1,27 @@
-class Rom
+class Room
+
+  attr_reader :category, :length, :width, :paint
 
   def initialize(category, length, width)
     @category = category
-    @length = 10
-    @width = width
+    @length = length.to_i
+    @width = width.to_i
+    @paint = false
   end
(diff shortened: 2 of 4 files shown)


KathleenScriver:
Would love to see more frequent commits.

lib/room.rb
```diff
@@ -1,12 +1,27 @@
-class Rom
+class Room
+
+  attr_reader :category, :length, :width, :paint
```
KathleenScriver, line 3: You'll want to be careful here, the interaction pattern did not ask for methods called #length or #width

lib/room.rb
```diff
@@ -1,12 +1,27 @@
...
+
+  def is_painted?
+    if @paint == false
+      false
+    elsif @paint == true
+      true
+    end
+  end
+
+  def paint
```
KathleenScriver, line 24: Your code has a syntax error which is causing problems. This might be where that's coming from. You have defined a method here called `paint`, but you also have a method called `paint` defined by your attr_reader. Ruby does not know what you want to do when you call `.paint`.

lib/room.rb
```diff
@@ -1,12 +1,27 @@
...
+    @width = width.to_i
+    @paint = false
   end
 
   def area
     @length * @width
   end
+
+  def is_painted?
+    if @paint == false
```
KathleenScriver, line 17: If the @paint instance variable is already holding a boolean, do we need to do all these comparisons to return the value of that instance variable? Think about how return values work to refactor this method.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def area
+    sq_footage = 0
+    self.rooms.each do |room|
+      sq_footage += room.area
+    end
+    sq_footage
+  end
+
+  def details
```
KathleenScriver, line 40: Here is another instance where you now have two methods called the same thing. This can confuse Ruby and cause unexpected behavior or syntax errors.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
+require '../lib/room'
+
+class House
+
+  attr_reader :price, :address, :rooms, :details
+
+  def initialize(price, address)
+    @price = price.tap { |str| str.delete!('$') }.to_i
```
KathleenScriver, line 8: Interesting approach to this. I think you could find quite a few string methods that would accomplish this for you though. Good job doing some research to get this accomplished though.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    false
```
KathleenScriver, line 18: You'll want to be sure to read all of the interaction pattern in the readme. It specified that this should only return false if the house is equal to or less than $500,000.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def rooms_from_category(category)
+    arr = []
+    self.rooms.each do |room|
```
KathleenScriver, line 23: `self` is not needed in this context.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+      if room.category == category
+        arr << room
+      else
```
KathleenScriver, line 26: You do not need an `else` if there are no other conditions you want to compare.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+    end
+    arr
```
KathleenScriver, line 29: Great. This is definitely a good way to get this one done using `#each`. I encourage you to look for other enumerables that will do this for you with much less code.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def area
+    sq_footage = 0
+    self.rooms.each do |room|
+      sq_footage += room.area
+    end
+    sq_footage
```
KathleenScriver, line 37: Good usage of `#each`

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def details
+    @details = {"price" => @price, "address" => @address}
```
KathleenScriver, line 41: I'm curious why you are saving this as an instance variable? Probably wouldn't want this stored as an instance variable since the data would need to be recalculated every time the price or address changed.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def price_per_square_foot
+    (self.price / self.area.to_f).round(2)
```
KathleenScriver, line 45: You typically do not need to use self in front of these methods. It is implied by Ruby. Instance methods can call other instance methods just by calling the name of the method.

KathleenScriver: Good job changing only one of the numbers to a float and rounding.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def rooms_sorted_by_area
+    self.rooms.sort_by {|room| room.area }
```
KathleenScriver, line 49: Great use of the `#sort_by` enumerable.

PR #21, opened 2020-05-13
Neeru Ram

neeruram1:
Independent Challenge #2

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,17 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize(price, address)
+    @price = price.delete("$").to_i
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room_param)
+    @rooms << room_param
+  end
+
+
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,15 +1,33 @@
 class House
   attr_reader :price,
               :address,
-              :rooms
+              :rooms,
+              :details
 
   def initialize(price, address)
     @price = price.delete("$").to_i
     @address = address
     @rooms = []
+    @details = {"price" => @price, "address" => @address}
   end
 
   def add_room(room_param)
     @rooms << room_param
+  end
+
+  def above_market_average?
+    @price > 500000
+  end
+
+  def rooms_from_category(category_param)
+    @rooms.find_all do |room|
+      room.category == category_param
+    end
+  end
+
+  def area
+    @rooms.sum do |room|
+      room.area
+    end
   end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -9,6 +9,7 @@ class House
     @address = address
     @rooms = []
     @details = {"price" => @price, "address" => @address}
+    @house_area = 0
   end
 
   def add_room(room_param)
@@ -26,8 +27,19 @@ class House
   end
 
   def area
-    @rooms.sum do |room|
+    @house_area = @rooms.sum do |room|
       room.area
     end
   end
+
+  def price_per_square_foot
+    price_per_square_foot = @price.to_f / @house_area.to_f
+    price_per_square_foot.to_f.round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.reverse! do |room|
+      room.area
+    end
+end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -12,6 +12,4 @@ class House
   def add_room(room_param)
     @rooms << room_param
   end
-
-
 end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -18,9 +18,4 @@ class Room
 def paint
   @paint = true
 end
-
-
-
-
-
 end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -1,12 +1,26 @@
-class Rom
+class Room
+  attr_reader :category
 
   def initialize(category, length, width)
     @category = category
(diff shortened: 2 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,45 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms,
+              :details
+
+  def initialize(price, address)
+    @price = price.delete("$").to_i
```
memcmahon, line 8: Great job with this `price` method

lib/house.rb
```diff
@@ -0,0 +1,45 @@
...
+    @address = address
+    @rooms = []
+    @details = {"price" => @price, "address" => @address}
```
memcmahon, line 11: This totally works, but might not be ideal.  duplicating storage of information in instance variables can make what are referred to as fat classes; so, this might be better as a method.

lib/house.rb
```diff
@@ -0,0 +1,45 @@
...
+    @house_area = 0
```
memcmahon, line 12: Similar to the `@details` variable above, this might be better as a method.  In programming, there is always a decision to be made on what is more expensive to your company or client: time or space.  With lots of instance variables, you are indicating that time is your most expensive resource (you would rather pay for the storage of information rather than wait for a time-expensive action to run).

lib/house.rb
```diff
@@ -0,0 +1,45 @@
...
+  def add_room(room_param)
+    @rooms << room_param
+  end
+
+  def above_market_average?
+    @price > 500000
+  end
+
+  def rooms_from_category(category_param)
+    @rooms.find_all do |room|
```
memcmahon, line 24: excellent use of find_all!

lib/house.rb
```diff
@@ -0,0 +1,45 @@
...
+      room.category == category_param
+    end
+  end
+
+  def area
+    @house_area = @rooms.sum do |room|
```
memcmahon, line 30: Great job with the sum method here, but we don't _need_ to assign this to a variable.

lib/house.rb
```diff
@@ -0,0 +1,45 @@
...
+      room.area
+    end
+  end
+
+  def price_per_square_foot
+    price_per_square_foot = @price.to_f / @house_area.to_f
+    price_per_square_foot.to_f.round(2)
+  end
```
memcmahon, line 38: Using this instance variable `@house_area` assumes that the method `area` will *always* be called before trying to determine price_per_square_foot, which may not be the case.  This method would work if you used `area` instead:

```ruby
def price_per_square_foot
    price_per_square_foot = @price.to_f / area.to_f
    price_per_square_foot.to_f.round(2)
  end
```

lib/house.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def rooms_sorted_by_area
+    @rooms.reverse! do |room|
```
memcmahon, line 41: So close!  But you would want to call a sort in there some where - reverse will just reorder the array it is called on.

```ruby
@rooms.sort_by do |room|
  room.area
end.reverse
```

lib/room.rb
```diff
@@ -1,12 +1,21 @@
...
     @length * @width
   end
+
+  def is_painted?
+    @paint == true
+  end
+
+def paint
+  @paint = true
+end
```
memcmahon, line 20: Watch the indentation here :)

test/house_test.rb
```diff
@@ -0,0 +1,124 @@
...
+end
+
+def test_it_has_rooms
+  house = House.new("$400000", "123 sugar lane")
+  assert_equal [], house.rooms
+end
+
+def test_it_can_add_rooms
+  house = House.new("$400000", "123 sugar lane")
+  assert_equal [], house.rooms
```
memcmahon, line 30: Since you have the test above, you don't _need_ to re-assert an empty array.

test/house_test.rb
```diff
@@ -0,0 +1,124 @@
...
+
+def test_its_below_market_average
+  house = House.new("$400000", "123 sugar lane")
+  assert_equal false, house.above_market_average?
+end
+
+def test_its_above_market_average
+  house = House.new("$600000", "123 sugar lane")
+  assert_equal true, house.above_market_average?
+end
```
memcmahon, line 48: Great job testing both outcomes!

test/house_test.rb
```diff
@@ -0,0 +1,124 @@
...
+  house.add_room(room_4)
+
+  assert_equal 1900, house.area
+end
+
+def test_it_has_details
+  house = House.new("$400000", "123 sugar lane")
+
+  assert_equal [400000, "123 sugar lane"], house.details.values
+  assert_equal ["price", "address"], house.details.keys
```
memcmahon, line 87: Good workaround for the hash assertion weirdness.  The way to get around that is to wrap your hash in parens:

```ruby
assert_equal ({price: "$400000", address: "123 sugar lane"}), house.details
```

test/house_test.rb
```diff
@@ -0,0 +1,124 @@
...
+  room_1 = Room.new(:bedroom, 10, '13')
+  room_2 = Room.new(:bedroom, 11, '15')
+  room_3 = Room.new(:living_room, 25, '15')
+  room_4 = Room.new(:basement, 30, '41')
+
+  house.add_room(room_1)
+  house.add_room(room_2)
+  house.add_room(room_3)
+  house.add_room(room_4)
+  house.area
```
memcmahon, line 102: This method should work without having to call the `area` method first.

PR #22, opened 2020-05-13
Chandler Hulstrom independent challenge

Chulstro:
Couldn't get through iteration 4. Got stuck on the sorting method

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -7,7 +7,7 @@ class House
               :rooms
 
   def initialize(cost, address)
-    @price = cost
+    @price = cost.delete("$").to_i
     @address = address
     @rooms = []
   end
@@ -17,7 +17,7 @@ class House
   end
 
   def above_market_average?
-    if @price.to_i >= 500000
+    if @price >= 500000
       true
     else
       false
@@ -41,5 +41,13 @@ class House
       area += room.area
     end
     area
+  end
+
+  def details
+    detail = {
+      "price" => @price,
+      "address" => @address
+    }
+    detail
   end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,14 @@
+require './lib/room'
+require 'pry'
+
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize(cost, address)
+    @price = cost
+    @address = address
+    @rooms = []
+  end
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -15,4 +15,31 @@ class House
   def add_room(room)
     @rooms << room
   end
+
+  def above_market_average?
+    if @price.to_i >= 500000
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(category)
+    cat_rooms = []
+    category = category
+    @rooms.each do |room|
+      if room.category == category
+        cat_rooms << room
+      end
+    end
+    cat_rooms
+  end
+
+  def area
+    area = 0
+    @rooms.each do |room|
+      area += room.area
+    end
+    area
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -50,4 +50,29 @@ class House
     }
     detail
   end
+
+  def price_per_square_foot
+    sqft_price = @price.to_f / self.area.to_f
+    sqft_price.round(2)
+  end
+
+  # def rooms_sorted_by_area
+  #   sorted_rooms =[]
+  #   @rooms.each do |room|
+  #     room = room
+  #     binding.pry
+  #     if sorted_rooms = []
+  #       sorted_rooms << room
+  #     elsif room.area > sorted_rooms[0].area
+  #       sorted_rooms.insert(0,room)
+  #     elsif room.area < sorted_rooms[0].area
+  #       sorted_rooms.insert(1, room)
+  #     elsif room.area > sorted_rooms[2].area
+  #       sorted_rooms.insert(2, room)
+  #     elsif room.area > sorted_rooms[3].area
+  #       sorted_rooms.insert(3, room)
+  #     end
+  #   end
+  #   sorted_rooms
+  # end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -11,4 +11,8 @@ class House
     @address = address
     @rooms = []
   end
+
+  def add_room(room)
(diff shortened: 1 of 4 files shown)


timomitchel:
Nice job committing throughout the challenge.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
+require './lib/room'
```
timomitchel, line 1: No need to require the room file up here. We only need it in our test files. We would only require this file if we needed to do a `Room.new` somewhere in the `House` class, which we don't.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+require 'pry'
+
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize(cost, address)
+    @price = cost.delete("$").to_i
```
timomitchel, line 10: Nice job with this effective and dynamic conversion!

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
```
timomitchel, line 15: Solid job here. A common pattern in challenges.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    @rooms << room
+  end
+
+  def above_market_average?
+    if @price >= 500000
```
timomitchel, line 20: Not quite. The `if` is redundant, and this should read `@price > 500000` based on the wording in the interaction pattern.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(category)
+    cat_rooms = []
+    category = category
```
timomitchel, line 29: No need to set this local variable. We have access to the param inside the block.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    @rooms.each do |room|
```
timomitchel, line 30: Nice job using `#each` to get the job done. This is also a great use case for `#find_all` which takes a condition and returns a subset of the original array for objects that meet that condition:
```
@rooms.find_all do |room|
room.category == category
end
```
So when the condition is true, we add that element to our new array, and if it is false, we disregard it.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+      if room.category == category
+        cat_rooms << room
+      end
+    end
+    cat_rooms
+  end
+
+  def area
+    area = 0
```
timomitchel, line 39: Again, nice job using `#each` to accomplish this. This is a great use case for `.sum`:
```
@rooms.sum do |room|
     room.area
end
```
Which does the totaling for you under the hood.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    @rooms.each do |room|
+      area += room.area
+    end
+    area
+  end
+
+  def details
+    detail = {
```
timomitchel, line 47: Great job building your hash here.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+      "price" => @price,
+      "address" => @address
+    }
+    detail
+  end
+
+  def price_per_square_foot
```
timomitchel, line 54: Nice job with this math. I like that you explicitly converted both of your integers to floats. We always want to have the same data types when we do math.

lib/house.rb
```diff
@@ -0,0 +1,78 @@
...
+    sqft_price = @price.to_f / self.area.to_f
+    sqft_price.round(2)
+  end
+
+  # def rooms_sorted_by_area
```
timomitchel, line 59: Yea, this is a bit more complex than it needs to be. This is a great use case for the `#sort_by` enumerable:
```
@rooms.sort_by do |room|
room.area
end.reverse
```
The order needs to be reversed to comply with the interaction pattern, but `#sort_by` will sort in ascending order

lib/room.rb
```diff
@@ -1,12 +1,15 @@
...
     @category = category
-    @length = 10
+    @length = length
     @width = width
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
   end
```
timomitchel, line 14: This class looks great up to this point, but you skipped right over the `is_painted?` and `paint` interaction.

test/house_test.rb
```diff
@@ -0,0 +1,82 @@
+require 'minitest/autorun'
+require './lib/house'
+require './lib/room'
+require 'pry'
+
+class HouseTest < Minitest::Test
+
+  def setup
```
timomitchel, line 8: Nice use of a setup method.

test/house_test.rb
```diff
@@ -0,0 +1,82 @@
...
+    assert_instance_of House, @house
+  end
+
+  def test_its_initial_values
+    assert_equal 400000, @house.price
+    assert_equal "123 sugar lane", @house.address
+    assert_equal [], @house.rooms
+  end
+
+  def test_rooms_exist
```
timomitchel, line 26: No need for this test, because this is the `House` test and we already tested this functionality in the `Room` test.

test/house_test.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def test_adding_rooms
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+
+    assert_equal [@room_1, @room_2], @house.rooms
+  end
+
+  def test_market_average_comparison
+    assert_equal false, @house.above_market_average?
```
timomitchel, line 39: It's always best to test both a true and a false scenario for boolean methods to make sure our method is working properly for both boolean literals.

test/house_test.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def test_rooms_categorically
```
timomitchel, line 42: Nice job with this test.

test/house_test.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def test_the_price_per_sqft
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+
+    assert_equal 210.53, @house.price_per_square_foot
+  end
```
timomitchel, line 72: Great tests here.

PR #24, opened 2020-05-13
Halverson real_estate_2005

dhalverson:
Assessment included here, I struggled a lot with the time allowed as I get in a hurry and then have issues. I am already planning on syncing up with my mentor and Graham on this content

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,10 @@
+class House
+  attr_reader :price,
+              :address
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+  end
+
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,10 +1,43 @@
 class House
   attr_reader :price,
-              :address
+              :address,
+              :rooms
 
   def initialize(price, address)
     @price = price
     @address = address
+    @rooms = []
+    @house_area = []
   end
 
+  def add_room(room_param)
+    @rooms << (room_param)
+  end
+
+# need to remove the $ from the string when converting to integer
+  # def above_market_average?
+  #   house.price.delete("$").to_i > 500000
+  # end
+
+  def rooms_from_category(category_param)
+    @rooms.select do |room|
+      room.category == (category_param)
+    end
+  end
+
+  def area
+    @rooms.each do |room|
+      @house_area << room.area
+    end
+    @house_area.sum
+  end
+
+# price needs to be converted to integer without $
+  def price_per_square_foot
+    @house_area.sum / house.price.to_i
+  end
+
+# use a sort method to order the room area's smallest to largest
+  # def rooms_sorted_by_area
+  #   @rooms.each do |room|
 end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -1,12 +1,22 @@
-class Rom
+class Room
+  attr_reader :category
 
   def initialize(category, length, width)
     @category = category
-    @length = 10
+    @length = length
     @width = width
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
   end
+
+  def is_painted?
+    false
+  end
+
+  # def paint
+  #   is_painted? = true
+  # end
+
 end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -15,8 +15,11 @@ class Room
     false
   end
 
-  # def paint
-  #   is_painted? = true
-  # end
+# this one confused me, I think I'm overthinking it
+  def paint
+    if is_painted? == false
+      true
+    end
+  end
 
 end
--- a/test/house_test.rb
+++ b/test/house_test.rb
@@ -1,0 +1,22 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/room'
+require './lib/house'
+
+class HouseTest < Minitest::Test
+
+  def test_it_exists
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_instance_of House, house
+  end
+
+  def test_it_has_attributes
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal "$400000", house.price
+    assert_equal "123 sugar lane", house.address
+  end
+
+
+end
--- a/test/house_test.rb
+++ b/test/house_test.rb
@@ -18,5 +18,93 @@ class HouseTest < Minitest::Test
     assert_equal "123 sugar lane", house.address
   end
 
+  def test_it_starts_with_zero_rooms
+    house = House.new("$400000", "123 sugar lane")
 
+    assert_equal [], house.rooms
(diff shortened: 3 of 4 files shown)


timomitchel:
Glad to hear you're planning to sync up with Graham and/or your mentor to work through the rest of this! I would have liked to see more frequent commits throughout this challenge.

lib/house.rb
```diff
@@ -0,0 +1,43 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
```
timomitchel, line 9: Nice setup up to this point, but no need for the `@house_area` instance variable.

lib/house.rb
```diff
@@ -0,0 +1,43 @@
...
+    @house_area = []
+  end
+
+  def add_room(room_param)
```
timomitchel, line 13: Well done here. A common pattern on challenges.

lib/house.rb
```diff
@@ -0,0 +1,43 @@
...
+    @rooms << (room_param)
```
timomitchel, line 14: No need for the parentheses

lib/house.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+# need to remove the $ from the string when converting to integer
```
timomitchel, line 17: I like this note to yourself.

lib/house.rb
```diff
@@ -0,0 +1,43 @@
...
+  # def above_market_average?
+  #   house.price.delete("$").to_i > 500000
+  # end
+
+  def rooms_from_category(category_param)
+    @rooms.select do |room|
```
timomitchel, line 23: Great job with this method but again no need for the parentheses.

lib/house.rb
```diff
@@ -0,0 +1,43 @@
...
+      room.category == (category_param)
+    end
+  end
+
+  def area
```
timomitchel, line 28: Nice job with this! A few things could improve it. This is a great use case for using the `.sum` enum:
```
@rooms.sum do |room|
      room.area
    end
```
Which will total the values as you iterate. If you're going to use `.each` then that's fine, but you want to set your accumulator as a local variable not an instance variable.

lib/house.rb
```diff
@@ -0,0 +1,43 @@
...
+    @house_area = []
+  end
+
+  def add_room(room_param)
+    @rooms << (room_param)
+  end
+
+# need to remove the $ from the string when converting to integer
+  # def above_market_average?
+  #   house.price.delete("$").to_i > 500000
```
timomitchel, line 19: Would have loved to see this `.delete("$").to_i` conversion in `initialize`. This is the correct implementation if you drop the `house.`

test/house_test.rb
```diff
@@ -0,0 +1,110 @@
...
+
+    assert_equal "$400000", house.price
+    assert_equal "123 sugar lane", house.address
+  end
+
+  def test_it_starts_with_zero_rooms
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal [], house.rooms
+  end
```
timomitchel, line 25: Testing looking solid here.

test/house_test.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  def test_it_can_add_rooms
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal [], house.rooms
```
timomitchel, line 32: No need for this assertion, as you already tested in the previous test.

test/house_test.rb
```diff
@@ -0,0 +1,110 @@
...
+## Iteration 3 ##
+
+  def test_it_is_above_market_average
+    skip
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal false, house.above_market_average?
+  end
+
+  def test_it_can_return_categories
```
timomitchel, line 48: Great job with this test setup and execution.

test/house_test.rb
```diff
@@ -0,0 +1,110 @@
...
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
+    assert_equal [room_4], house.rooms_from_category(:basement)
+  end
+
+  def test_it_can_get_area
```
timomitchel, line 64: Again, flawless test.

test/house_test.rb
```diff
@@ -0,0 +1,110 @@
...
+    house = House.new("$400000", "123 sugar lane")
+
+    house.add_room(room_4)
+    house.add_room(room_1)
+    house.add_room(room_3)
+    house.add_room(room_2)
+
+    house.area
+
+    assert_equal 210.53, house.price_per_square_foot
```
timomitchel, line 93: Nice job here, but this method should be able to function apart from the `house.area` call.

test/house_test.rb
```diff
@@ -0,0 +1,110 @@
...
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house = House.new("$400000", "123 sugar lane")
+
+    house.add_room(room_4)
+    house.add_room(room_1)
+    house.add_room(room_3)
+    house.add_room(room_2)
+
+    assert_equal [room_1, room_2, room_3, room_4], house.rooms_sorted_by_area
```
timomitchel, line 108: This setup should be in the opposite order with room_4 at the front of the array according to the interaction pattern.

lib/room.rb
```diff
@@ -1,12 +1,25 @@
...
+    @length = length
     @width = width
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
   end
+
+  def is_painted?
```
timomitchel, line 14: We almost never want to hardcode a value like this, because it's not dynamic enough.

lib/room.rb
```diff
@@ -1,12 +1,25 @@
...
+    false
+  end
+
+# this one confused me, I think I'm overthinking it
```
timomitchel, line 18: Totally, as I said this was a tough one. Remember that instance variables are tools that you can use when multiple methods of the same class need to access or change a value.

test/room_test.rb
```diff
@@ -3,22 +3,39 @@
...
     assert_equal 180, room2.area
   end
+
+  def test_it_is_painted
+    room = Room.new(:bedroom, 10, '13')
+
+    assert_equal false, room.is_painted?
+  end
+
+  def test_it_can_be_painted
```
timomitchel, line 33: The test here is exactly right.

PR #26, opened 2020-05-13
Real_estate_2005_KB

Kathybui732:
Kathy Bui

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -26,4 +26,11 @@ class House
       area + room.area
     end
   end
+
+  def details
+    details = {
+      "price" => @price.delete("$").to_i,
+      "address" => @address
+    }
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -10,4 +10,14 @@ class House
   def add_room(room)
     @rooms << room
   end
+
+  def above_market_average?
+    @price >= "$500000"
+  end
+
+  def rooms_from_category(room_cat)
+    @rooms.sort_by do |room|
+      room.category == room_cat
+    end
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -44,4 +44,10 @@ class House
     end
     room_areas.reverse
   end
+
+  def rooms_by_category
+    @rooms.group_by do |room|
+      room.category
+    end
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -37,4 +37,11 @@ class House
   def price_per_square_foot
     (@price.delete("$").to_f / area).round(2)
   end
+
+  def rooms_sorted_by_area
+    room_areas = @rooms.sort_by do |room|
+      room.area
+    end
+    room_areas.reverse
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -33,4 +33,8 @@ class House
       "address" => @address
     }
   end
+
+  def price_per_square_foot
+    (@price.delete("$").to_f / area).round(2)
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -6,4 +6,8 @@ class House
     @address = address
     @rooms = []
   end
+
+  def add_room(room)
+    @rooms << room
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,9 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -16,8 +16,14 @@ class House
   end
 
   def rooms_from_category(room_cat)
-    @rooms.sort_by do |room|
+    @rooms.select do |room|
       room.category == room_cat
+    end
+  end
+
+  def area
+    @rooms.reduce(0) do |area, room|
+      area + room.area
     end
   end
 end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -5,9 +5,18 @@ class Room
     @category = category
     @length = length
     @width = width
+    @painted = false
   end
 
   def area
     @length * @width.to_i
+  end
+
+  def is_painted?
+    @painted
+  end
+
+  def paint
+    @painted = true
   end
 end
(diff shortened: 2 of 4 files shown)


KathleenScriver:
Great job with all the commits!

test/room_test.rb
```diff
@@ -9,16 +9,30 @@ def test_it_exists
...
     room1 = Room.new(:bedroom, 10, '13')
     room2 = Room.new(:living_room, 15, '12')
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_is_it_painted
+    # skip
+    room = Room.new(:bedroom, 10, '13')
+    refute room.is_painted?
```
KathleenScriver, line 28: Good use of refute here. I would just recommend naming the test something slightly more descriptive. We are actually testing that the room starts unpainted. OR, if the test is called `test_is_it_painted`, you could even test the initial state of is_painted? being false, then paint it, then test that is_painted? is true all in one test.

test/house_test.rb
```diff
@@ -0,0 +1,140 @@
...
+class HouseTest < Minitest::Test
+
+  def test_it_exists
+    house = House.new("$400000", "123 sugar lane")
+    assert_instance_of House, house
+  end
+
+  def test_it_has_a_price
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal "$400000", house.price
```
KathleenScriver, line 15: Be careful here. You'll want to read and reread the interaction pattern. This method should return an integer, not a string.

test/house_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    house.add_room(room_2)
+    assert_equal [room_1, room_2], house.rooms
+  end
+
+  def test_it_is_only_above_market_price_if_over_500000
+    # skip
+    house = House.new("$400000", "123 sugar lane")
+    house_2 = House.new("$500000", "234 sugar lane")
+    refute house.above_market_average?
+    assert house_2.above_market_average?
```
KathleenScriver, line 42: Great job testing both the `true` and `false` cases for this method.

test/house_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+    assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
+    assert_equal [room_4], house.rooms_from_category(:basement)
+    assert_equal [room_3], house.rooms_from_category(:living_room)
```
KathleenScriver, line 58: Great job testing these cases. Keep in mind though, if we had another 5 or 6 categories, it would not be necessary to test every single category, maybe just 2 would be sufficient to feel confident that the method is doing what it needs to do.

test/house_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+    expected = {
+      "price" => 400000,
+      "address" => "123 sugar lane"
+    }
+    assert_equal expected, house.details
```
KathleenScriver, line 90: Great job testing the hash here.

test/house_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    house.add_room(room_3)
+    house.add_room(room_4)
+    expected = {
+      :bedroom => [room_1, room_2],
+      :living_room => [room_3],
+      :basement => [room_4]
+    }
+    assert_equal expected, house.rooms_by_category
+  end
+end
```
KathleenScriver, line 140: Great tests. Very thorough and specific.

lib/room.rb
```diff
@@ -1,12 +1,22 @@
-class Rom
+class Room
+  attr_reader :category
```
KathleenScriver, line 2: Great job only putting `category` into the attr_reader. We didn't want length or width to be methods exposed outside the class.

lib/room.rb
```diff
@@ -1,12 +1,22 @@
...
 
-  def initialize(category, length, width)
+  def initialize(category, length = 10, width)
```
KathleenScriver, line 4: Interesting choice to set a default value here. Wondering why you included that. The interaction patterns did not ask for a default value. AND, think about this. If I did this: Room.new(:bedroom, 50), where I wanted the length to be 10 and the width to be 50, what would happen in this scenario? Default methods should always be the very last argument.

lib/house.rb
```diff
@@ -0,0 +1,53 @@
...
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    @price >= "$500000"
+  end
+
+  def rooms_from_category(room_cat)
+    @rooms.select do |room|
```
KathleenScriver, line 19: Great use of the find_all enumerable.

lib/house.rb
```diff
@@ -0,0 +1,53 @@
...
+      room.category == room_cat
+    end
+  end
+
+  def area
+    @rooms.reduce(0) do |area, room|
```
KathleenScriver, line 25: Great job using reduce for this. I also love the naming of your accumulator, very specific and descriptive.

lib/house.rb
```diff
@@ -0,0 +1,53 @@
...
+      area + room.area
+    end
+  end
+
+  def details
+    details = {
```
KathleenScriver, line 31: Curious why you are storing this in a variable? Is that needed? Try it without and see!

lib/house.rb
```diff
@@ -0,0 +1,53 @@
...
+      "address" => @address
+    }
+  end
+
+  def price_per_square_foot
+    (@price.delete("$").to_f / area).round(2)
+  end
+
+  def rooms_sorted_by_area
+    room_areas = @rooms.sort_by do |room|
```
KathleenScriver, line 42: Great job using this sort_by enumerable for this use-case.

lib/house.rb
```diff
@@ -0,0 +1,53 @@
...
+      room.area
+    end
+    room_areas.reverse
+  end
+
+  def rooms_by_category
+    @rooms.group_by do |room|
```
KathleenScriver, line 49: This works, great job! Group_by can be helpful, I'd challenge you to explore other ways to accomplish this same task to get experience using a few other enumerables.

PR #27, opened 2020-05-13
Iterations 1 - 4!

Strompy:
Made it through all iterations. 🤘

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,11 +1,12 @@
 require './lib/room'
 
 class House
-  attr_reader :price, :address
+  attr_reader :price, :address, :rooms 
 
   def initialize(price, address)
     @price = price
     @address = address
+    @rooms = []
   end
-  
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,16 +1,21 @@
 require './lib/room'
 
 class House
-  attr_reader :price, :address, :rooms
+  attr_reader :price, :address, :rooms, :value
 
   def initialize(price, address)
     @price = price
     @address = address
     @rooms = []
+    @value = @price.delete("$").to_i
   end
 
   def add_room(room)
-    @rooms << room 
+    @rooms << room
+  end
+
+  def above_market_average?
+    @value > 500000
   end
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -18,4 +18,10 @@ class House
     @value > 500000
   end
 
+  def rooms_from_category(category)
+    @rooms.find_all do |room|
+      category == room.category
+    end
+  end
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,4 +1,11 @@
 require './lib/room'
 
 class House
-end 
+  attr_reader :price, :address
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+  end
+  
+end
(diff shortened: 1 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,64 @@
+require './lib/room'
```
memcmahon, line 1: You don't need to require the room file here - only require files when you need that class definition (if you were calling Room.new, for example)

lib/house.rb
```diff
@@ -0,0 +1,64 @@
...
+
+class House
+  attr_reader :price, :address, :rooms, :value
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+    @value = @price.delete("$").to_f
```
memcmahon, line 10: We acually wanted a method `price` to return the integer :)

lib/house.rb
```diff
@@ -0,0 +1,64 @@
...
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    @value > 500000
+  end
+
+  def rooms_from_category(category)
+    @rooms.find_all do |room|
```
memcmahon, line 22: excellent use of find_all!

lib/house.rb
```diff
@@ -0,0 +1,64 @@
...
+      category == room.category
+    end
+  end
+
+  def area
+    house_area = 0
+    @rooms.each do |room|
```
memcmahon, line 29: Great use of `each` - you could also look into the method `sum`

lib/house.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def details
+    house_details = Hash.new
+    house_details["price"] = @value
+    house_details["address"] = @address
+    house_details
+  end
+
+  def price_per_square_foot
+    psf = @value / area
```
memcmahon, line 43: Try not to use acronyms as variable names - will you remember in 3 days what `psf` means? (if you weren't looking at the code)

lib/house.rb
```diff
@@ -0,0 +1,64 @@
...
+    psf.round(2)
+  end
+
+  def rooms_sorted_by_area
+    rooms.sort {|a,b| b.area <=> a.area}
```
memcmahon, line 48: Similar to my note above, single character variables are not great - room_a and room_b would be better.

lib/house.rb
```diff
@@ -0,0 +1,64 @@
...
+    cat_sort = Hash.new
+
+    bedrooms = rooms.find_all {|room| room.category == :bedroom}
+    cat_sort[:bedroom] = bedrooms
+    living_rooms = rooms.find_all {|room| room.category == :living_room}
+    cat_sort[:living_room] = living_rooms
+    basements = rooms.find_all {|room| room.category == :basement}
+    cat_sort[:basement] = basements
+    cat_sort
+    # not dynamic yet but works for iteration
```
memcmahon, line 61: Yea.... works for the iteration, but not fully functional.

test/house_test.rb
```diff
@@ -0,0 +1,121 @@
...
+
+class HouseTest < Minitest::Test
+  def test_it_exists
+    house = House.new("$400000", "123 sugar lane")
+    assert_instance_of House, house
+  end
+
+  def test_it_has_attributes
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal "$400000", house.price
```
memcmahon, line 14: Keep a close eye on the interaction pattern - we wanted `price` to return an integer!

PR #28, opened 2020-05-13
James Belta

JBelta:


test/room_test.rb
```diff
@@ -9,16 +9,22 @@ def test_it_exists
...
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_is_it_painted?
+    room = Room.new(:bedroom, 10, '13')
+    assert_equal false, room.is_painted?
+
+    room.paint
+    assert_equal true, room.is_painted?
```
KathleenScriver, line 28: Great job testing the initial state of is_painted?, applying a method, and then testing the resulting state of is_painted? That's a great pattern.

test/house_test.rb
```diff
@@ -0,0 +1,66 @@
...
+    room_2 = Room.new(:bedroom, 11, '15')
+    house.add_room(room_1)
+    house.add_room(room_2)
+
+    assert_equal [room_1, room_2], house.rooms
+  end
+
+  def test_above_market_average?
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal false, house.above_market_average?
```
KathleenScriver, line 32: When you have a method that returns a boolean, you'll want to test both the `true` case and the `false` case, even if the interaction pattern doesn't explicitly ask for that.

test/house_test.rb
```diff
@@ -0,0 +1,66 @@
...
+      room_3 = Room.new(:living_room, 25, '15')
+      room_4 = Room.new(:basement, 30, '41')
+      house.add_room(room_1)
+      house.add_room(room_2)
+      house.add_room(room_3)
+      house.add_room(room_4)
+
+      assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
+      assert_equal [room_4], house.rooms_from_category(:basement)
+      assert_equal ({"price" => 400000, "address" => "123 sugar lane"}), house.details
```
KathleenScriver, line 48: Each method should have its own test.

KathleenScriver: Great job figuring out how to do an assertion with a hash in minitest.

test/house_test.rb
```diff
@@ -0,0 +1,66 @@
...
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal 210.53, house.price_per_square_foot
+    assert_equal [room_4, room_3, room_2, room_1], house.rooms_sorted_by_area
+    assert_equal ({:bedroom => [room_1, room_2], :living_room => room_3, :basement => room_4}), house.rooms_by_category
```
KathleenScriver, line 64: Same thing here, each method should have its own test.

KathleenScriver, line 64: Also, maybe the interaction pattern wasn't super clear, but the values of each key in this hash should be an array, even if that array only has one element in it. Typically we would see that consistency across the hash.

lib/room.rb
```diff
@@ -1,12 +1,23 @@
-class Rom
+class Room
+  attr_reader :category, :length, :width
```
KathleenScriver, line 2: You'll just want to be careful with adding all attributes to the attr_reader. The interaction pattern did not ask for methods called `length` or `width`.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
+require './lib/room'
+
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price.delete("$").to_i
```
KathleenScriver, line 7: Great job figuring out a string method to complete this one.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def add_room(room)
+    @rooms << room
+  end
+  def above_market_average?
+    @price > 500000
+  end
+
+  def rooms_from_category(category)
+    @rooms.find_all do |room|
```
KathleenScriver, line 20: Great job using the appropriate enumerable here.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+      category == room.category
+    end
+  end
+
```
KathleenScriver, line 24: Be sure to read the interaction pattern carefully. It looks like you missed a method for the House class called `#area` which would have helped you in some of the methods below.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+  def details
+    {"price" => @price, "address" => @address}
+  end
+
+  def price_per_square_foot
+    total_sqft = @rooms.sum do |room|
+      sum = 0
+      sum += (room.length * room.width.to_i)
+    end
+    total = @price.to_f / total_sqft.to_f
```
KathleenScriver, line 34: Great job turning the integers to floats. In Ruby though, only one of the numbers in division needs to be a float, either the numerator or the denominator.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+    total.round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort { |room| room.area}
```
KathleenScriver, line 39: So so close!!! You'll want to investigate the differences between sort and sort_by. Play in a pry session to see the differences.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def rooms_by_category
+    #room_hash = Hash.new(0)
```
KathleenScriver, line 43: Having this accumulator out here would have been a good idea, probably don't need a default value though. I encourage you to keep playing with this one!

PR #29, opened 2020-05-13
Taija Warbelow

twarbelow:


lib/room.rb
```diff
@@ -1,12 +1,25 @@
-class Rom
+class Room
+  attr_reader :category,
+              :length,
+              :width,
+              :painted
```
memcmahon, line 5: Since we have the method `is_painted?` below, we do not also need an attr_reader ;)

lib/room.rb
```diff
@@ -1,12 +1,25 @@
...
+    length * width.to_i
+  end
+
+  def is_painted?
+    painted
+  end
+
+  def paint
+    @painted = true
   end
```
memcmahon, line 24: Great job with this whole paint thing!

lib/house.rb
```diff
@@ -0,0 +1,36 @@
...
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
```
memcmahon, line 15: Looks like you missed the method `.area`

lib/house.rb
```diff
@@ -0,0 +1,36 @@
...
+  def price_per_square_foot
+    total_square_feet = 0
+    rooms.each do |room|
+      total_square_feet += room.area.to_f
+    end
+    (price.tr("$", "").to_f / total_square_feet).round(2)
+  end
```
memcmahon, line 22: Great job using `each` here; you could also look into the method `.sum`

lib/house.rb
memcmahon: Actually, on second look, it seems like you may have skipped all of iteration 3?

lib/house.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def sort_rooms_by_area #not working
+    return rooms.sort_by(&:area)
+    # rooms.sort_by { |room| room.area }
+  end
```
memcmahon, line 27: This code here is working - see my note on your test, but the method is not named correctly; we asked for `rooms_sorted_by_area`

lib/house.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def rooms_by_category #not working
+    sorted = {}
+    rooms.map do |room|
+      sorted[room.category] = room
+    end
+    return sorted
+  end
```
memcmahon, line 35: You're on the right track here, but map is not quite what you want.  `group_by` might work better

test/house_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    @house = House.new("$400000", "123 sugar lane")
+    @room_1 = Room.new(:bedroom, 10, '13')
+    @room_2 = Room.new(:bedroom, 11, '15')
+    @room_3 = Room.new(:living_room, 25, '15')
+    @room_4 = Room.new(:basement, 30, '41')
+  end
+
+  def test_it_has_price
+    @house = House.new("$400000", "123 sugar lane")
+    assert_equal "$400000", @house.price
```
memcmahon, line 17: Keep an eye on the interaction pattern - we wanted `price` to return an integer

test/house_test.rb
```diff
@@ -0,0 +1,51 @@
...
+  def test_it_has_price_per_square_foot
+    @house.add_room(@room_4)
+    @house.add_room(@room_1)
+    @house.add_room(@room_3)
+    @house.add_room(@room_2)
+    assert_equal 210.53, @house.price_per_square_foot
+  end
+
+  def test_it_can_sort_rooms_by_area
+    assert_equal [@room_1, @room_2, @room_3, @room_4], @house.sort_rooms_by_area
```
memcmahon, line 41: You need to add rooms to your house in order for this test to work 😉

PR #30, opened 2020-05-13
Austin Aspaas

evilaspaas1:


timomitchel:
Would have liked to see more committing and properly formatted commit messages throughout!

lib/house.rb
```diff
@@ -0,0 +1,21 @@
+require "pry"
+
+class House
+  attr_reader :price,
+              :address,
+              :rooms
```
timomitchel, line 6: Looks like you missed a few commas here. And, there was no need to create a `:room_1` or `:room_2` `attr_reader` here.

lib/house.rb
```diff
@@ -0,0 +1,21 @@
...
+              :room_1
+              :room_2
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
```
timomitchel, line 12: Nice job with this initialize setup!

lib/house.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def add_room(room_1, room_2)
```
timomitchel, line 15: Not quite. Based on the interaction pattern, we only pass one room object at a time. So, we only define one room parameter for this method.

lib/room.rb
```diff
@@ -1,12 +1,25 @@
...
+attr_reader :bedroom
   def initialize(category, length, width)
     @category = category
     @length = 10
     @width = width
+    @is_painted = false
+  end
+
+  def category
+    @category = :bedroom
```
timomitchel, line 11: Not quite what we want here. We wanted this to be an `attr_reader` for `@category` that is  set dynamically via `initialize`.

lib/room.rb
```diff
@@ -1,12 +1,25 @@
...
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_painted?
```
timomitchel, line 18: We only want to return the value of our instance variable here! This is so close!

lib/room.rb
```diff
@@ -1,12 +1,25 @@
...
+    @is_painted = false
+  end
+
+  def paint
```
timomitchel, line 22: Nice job here.

lib/room.rb
```diff
@@ -1,12 +1,25 @@
-class Rom
-
+class Room
+attr_reader :bedroom
```
timomitchel, line 2: No need for this `attr_reader` as it does not have a corresponding instance variable and we never call `.bedroom`

test/room_test.rb
```diff
@@ -1,24 +1,31 @@
...
+    room = Room.new(:bedroom, 10, '13')
+    assert_equal 130, room.area
+  end
+
+  def test_is_painted?
+    room = Room.new(:bedroom, 10, '13')
+    assert_equal false, room.is_painted?
+  end
+
+  def test_paint
```
timomitchel, line 27: Close! But, the interaction pattern is not concerned with the return value of the `#paint` method, but instead concerned that `#paint` changes the value of `@is_painted`. You were on the right track with this interaction!

test/house_test.rb
```diff
@@ -0,0 +1,34 @@
...
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal "123 sugar lane", house.address
+  end
+
+  def test_it_has_rooms
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal [], house.rooms
+  end
+
+  def test_it_can_add_room
```
timomitchel, line 28: Close! But, we need to call our `#add_room` method here!

test/house_test.rb
```diff
@@ -0,0 +1,34 @@
...
+require "./lib/house"
+require "pry"
+
+class HouseTest < Minitest::Test
+  def test_it_exists
+    house = House.new(400000, "123 sugar lane")
+    assert_instance_of House, house
+  end
+
+  def test_it_has_price
```
timomitchel, line 13: Nice job testing your attributes!

PR #31, opened 2020-05-13
Dan Sehl 2005BE

dtsehl87:
This was a fun one!

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -7,6 +7,7 @@ class House
     @address = address
     @rooms = []
     @details = {}
+    @room_categories = {}
   end
 
   def add_room(room)
@@ -48,6 +49,12 @@ class House
   def rooms_sorted_by_area
     @rooms.sort do |room|
       room.area
+    end
+  end
+
+  def rooms_by_category
+    @rooms.group_by do |room|
+      room.category
     end
   end
 
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,15 @@
+class House
+
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -41,5 +41,14 @@ class House
     @details
   end
 
+  def price_per_square_foot
+    (@price.gsub('$', '').to_f / self.area).round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort do |room|
+      room.area
+    end
+  end
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -6,10 +6,40 @@ class House
     @price = price
     @address = address
     @rooms = []
+    @details = {}
   end
 
   def add_room(room)
     @rooms << room
   end
+
+  def above_market_average?
+    if @price.gsub('$', '').to_i >= 500_000
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(category)
+    @rooms.find_all do |room|
+      room.category == category
+    end
+  end
+
+  def area
+    result = 0
+    @rooms.each do |room|
+      result += room.area
+    end
+    result
+  end
+
+  def details
+    @details["price"] = @price.gsub('$', '').to_i
+    @details["address"] = @address
+    @details
+  end
+
 
 end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -1,12 +1,23 @@
-class Rom
+class Room
+
+  attr_reader :category
 
   def initialize(category, length, width)
     @category = category
-    @length = 10
+    @length = length
     @width = width
+    @painted = false
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_painted?
+    @painted
+  end
+
+  def paint
+    @painted = true
   end
 end
--- a/test/house_test.rb
+++ b/test/house_test.rb
@@ -90,4 +90,47 @@ class HouseTest < Minitest::Test
     assert_equal ({"price" => 400000, "address" => "123 sugar lane"}), house.details
   end
 
+  def test_house_can_have_price_per_square_foot
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal 210.53, house.price_per_square_foot
+  end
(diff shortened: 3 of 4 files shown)


timomitchel:
Nice job committing throughout, but those commit messages could use some work!

lib/house.rb
```diff
@@ -0,0 +1,61 @@
+class House
+
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price
```
timomitchel, line 6: We were hoping to see the string conversion to an integer here in initialize.

lib/house.rb
```diff
@@ -0,0 +1,61 @@
...
+    @address = address
+    @rooms = []
+    @details = {}
+    @room_categories = {}
```
timomitchel, line 10: No need for either of these instance variables based on the interaction pattern.

lib/house.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def add_room(room)
```
timomitchel, line 13: Nice, this is exactly what we wanted and a common pattern you will see in challenges.

lib/house.rb
```diff
@@ -0,0 +1,61 @@
...
+    @rooms << room
+  end
+
+  def above_market_average?
+    if @price.gsub('$', '').to_i >= 500_000
```
timomitchel, line 18: Not quite. We wanted our condition to read: `@price.gsub('$', '').to_i > 500_000` based on the wording of the interaction pattern. And, the `if` is redundant, because the above will always return a true or false boolean literal when the condition runs

lib/house.rb
```diff
@@ -0,0 +1,61 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(category)
+    @rooms.find_all do |room|
```
timomitchel, line 26: The perfect enumerable for the job here!

lib/house.rb
```diff
@@ -0,0 +1,61 @@
...
+      room.category == category
+    end
+  end
+
+  def area
```
timomitchel, line 31: Nice job using `#each` effectively here. This is also the perfect use case for the `#sum` enumerable:
```
@rooms.sum do |room|
room.area
end
```

lib/house.rb
```diff
@@ -0,0 +1,61 @@
...
+    result = 0
+    @rooms.each do |room|
+      result += room.area
+    end
+    result
+  end
+
+  def details
+    @details["price"] = @price.gsub('$', '').to_i
```
timomitchel, line 40: Notice how you're repeating yourself doing the integer conversion in each method. This is why we want to do that conversion in initialize, so we only do it once.

lib/house.rb
```diff
@@ -0,0 +1,61 @@
...
+    @details["address"] = @address
+    @details
```
timomitchel, line 42: Without the instance variable and with the conversion already run this method could simply read:
{"price" => @price, "address" => @address}

lib/house.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def price_per_square_foot
+    (@price.gsub('$', '').to_f / self.area).round(2)
```
timomitchel, line 46: Nice job with this method.

lib/house.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def rooms_sorted_by_area
```
timomitchel, line 49: Close, but according to the interaction pattern we wanted these in reverse order, so we needed to reverse the result of our `sort` block with a `.reverse`.

lib/house.rb
```diff
@@ -0,0 +1,61 @@
...
+    @rooms.sort do |room|
+      room.area
+    end
+  end
+
+  def rooms_by_category
```
timomitchel, line 55: Great job using the correct enum for the job here.

lib/room.rb
```diff
@@ -1,12 +1,23 @@
...
+    @length * @width.to_i
+  end
+
+  def is_painted?
+    @painted
+  end
+
+  def paint
+    @painted = true
   end
```
timomitchel, line 22: Nice job with these methods.

test/house_test.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  def test_it_can_have_rooms_addedd
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    assert_equal [room_1, room_2], house.rooms
+  end
```
timomitchel, line 34: Really nice job with this testing.

test/house_test.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  def test_it_is_avbove_market_average_if_500k_and_avove
+    house_1 = House.new("$400000", "123 sugar lane")
+
+    assert_equal false, house_1.above_market_average?
+
+    house_2 = House.new("$500000", "321 apple drive")
+
+    assert_equal true, house_2.above_market_average?
```
timomitchel, line 43: So close, but this should return a false. Still, I like the robust testing of this method.

test/house_test.rb
```diff
@@ -0,0 +1,135 @@
...
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
```
timomitchel, line 61: Great test.

test/house_test.rb
```diff
@@ -0,0 +1,135 @@
...
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal ({"price" => 400000, "address" => "123 sugar lane"}), house.details
```
timomitchel, line 90: Nice job figuring out how to pass a hash to your `#assert_equal` method.

test/house_test.rb
```diff
@@ -0,0 +1,135 @@
...
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal [room_4, room_3, room_2, room_1], house.rooms_sorted_by_area
```
timomitchel, line 118: The test setup is correct here.

timomitchel:
Correction, the `rooms_sorted_by_area` was implemented correctly. You get credit for that method.

PR #32, opened 2020-05-13
Dani Coleman

dcoleman21:
Still working on learning these concepts. Really frustrating.

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,22 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    if @price > $500000
+      true
+    else
+      false
+    end
+  end
+end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -1,12 +1,18 @@
-class Rom
-
+class Room
+  attr_reader :category
+              :length
+              :width
   def initialize(category, length, width)
     @category = category
-    @length = 10
+    @length = length
     @width = width
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_painted?
+    
   end
 end
--- a/test/house_test.rb
+++ b/test/house_test.rb
@@ -1,0 +1,28 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/room'
+require './lib/house'
+
+class HouseTest < Minitest::Test
+  def test_it_exists
+    house = House.new("$400000", "123 sugar lane")
+    assert_instance_of House, house
+  end
+
+  def test_it_can_add_rooms
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    assert_equal [room_1, room_2], house.rooms
+  end
+
+  def test_is_is_above_market_average?
+
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal false, house.above_market_average?
+  end
+end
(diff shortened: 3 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,22 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
```
timomitchel, line 9: Nice job here. Exactly the setup of instance variables and `attr_reader`s we were looking for.

lib/house.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def add_room(room)
```
timomitchel, line 11: Yes, this is what we wanted, and a common pattern you will see in challenges.

lib/house.rb
```diff
@@ -0,0 +1,22 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms
+  def initialize(price, address)
+    @price = price
```
timomitchel, line 6: We needed to convert this `price` argument to an integer in order to do our comparison in `#above_market_average`.

lib/house.rb
```diff
@@ -0,0 +1,22 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    if @price > $500000
```
timomitchel, line 16: Close, but not quite. 
```
[2] pry(main)> "400000" > $500000
ArgumentError: comparison of String with nil failed
from (pry):2:in `>'
```
That's from my terminal, and I imagine is what your tests were saying. We need integers or floats to do comparisons this way. And, the `if` is redundant here. If price has been converted to an integer, then we only need:
```
@price > $500000
```
which will always return either a true or false.

lib/room.rb
```diff
@@ -1,12 +1,18 @@
-class Rom
-
+class Room
+  attr_reader :category
+              :length
+              :width
   def initialize(category, length, width)
     @category = category
-    @length = 10
+    @length = length
```
timomitchel, line 7: Nice job making this change from the hard coded value.

lib/room.rb
```diff
@@ -1,12 +1,18 @@
...
     @width = width
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
```
timomitchel, line 12: This works, but it's better if we do the `.to_i` conversion in initialize.

lib/room.rb
```diff
@@ -1,12 +1,18 @@
...
+  end
+
+  def is_painted?
```
timomitchel, line 15: Looks like you were on the right track. As I showed in the demo we needed to use an instance variable `@is_painted` and return it via this method.

test/house_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def test_it_can_add_rooms
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    assert_equal [room_1, room_2], house.rooms
```
timomitchel, line 19: Great test.

test/house_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def test_is_is_above_market_average?
+
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal false, house.above_market_average?
```
timomitchel, line 26: Nice test setup here as well.

test/room_test.rb
```diff
@@ -9,16 +9,22 @@ def test_it_exists
...
   def test_it_can_get_area
-    skip
     room1 = Room.new(:bedroom, 10, '13')
     room2 = Room.new(:living_room, 15, '12')
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_if_room_is_painted
+    room = Room.new(:bedroom, 10, '13')
```
timomitchel, line 24: The tests are looking good!

PR #33, opened 2020-05-13
Sienna Kopf independent challenge 2 real_estate_2005

sienna-kopf:


lib/house.rb
```diff
@@ -0,0 +1,91 @@
+class House
+  attr_reader :price,
+              :address,
+              :rooms,
+              :categorical_rooms,
+              :total_area
```
memcmahon, line 6: Since these are method below, they do not need to be attr_readers

lib/house.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    return true if price > 500000
+    false
+  end
```
memcmahon, line 21: This is great - you could also rely on the return value of the condition:

```ruby
def above_market_average?
    price > 500000
end
```

lib/house.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def rooms_from_category(category)
+    categorical_rooms = rooms.select do |room|
+      room.category == category
+    end
```
memcmahon, line 26: You don't need to store the return value of select in a variable

lib/house.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def area
+    total_area = 0
+    rooms.each do |room|
```
memcmahon, line 31: Great use of `each` - you could also look into the `sum` method :)

lib/house.rb
```diff
@@ -0,0 +1,91 @@
...
+      "address" => address
+    }
+  end
+
+  def price_per_square_foot
+    (price.to_f / area).round(2)
+  end
+
+  def rooms_sorted_by_area
+    rooms_by_area = @rooms.sort_by {|room| room.area}.reverse!
```
memcmahon, line 49: great use of `sort_by` and `reverse`!

lib/house.rb
```diff
@@ -0,0 +1,91 @@
...
+    # # rooms_by_category.each do |categories, rooms|
+    # #   rooms.each do |room|
+    # #     if room.category == categories
+    # #       rooms_by_category.push "room"
+    # #     end
+    # #   end
+    # # end
+    # #
+    # # rooms_by_category
+  end
```
memcmahon, line 88: You are definitely on the right track here!  A few nudges in the right direction:

remember you have access to the `rooms_from_category` method

Look into `group_by`

test/house_test.rb
```diff
@@ -0,0 +1,102 @@
...
+    @house.add_room (@room_2)
+    assert_equal [@room_1, @room_2], @house.rooms
+  end
+
+  def test_it_is_above_market_average
+    assert_equal false, @house.above_market_average?
+
+    expensive_house = House.new("$600000", '987 strawberry fields ave')
+    assert_equal true, expensive_house.above_market_average?
+  end
```
memcmahon, line 40: Great job testing both scenarios!

PR #34, opened 2020-05-13
Eric Hale

EHale64:
Wasn't too excited to do this with pry not working but I made it!

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -14,4 +14,12 @@ class House
   def add_room(room)
     @rooms << room
   end
+
+  def above_market_average?
+    price > 500000
+  end
+
+  def rooms_from_category(category)
+    @rooms.find_all { |room| room.category == category}
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -37,4 +37,8 @@ class House
   def price_per_square_foot
     (price / area.to_f).round(2)
   end
+
+  def rooms_sorted_by_area
+    @rooms.sort { |room| room.area}
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,14 @@
+class House
+  attr_reader :address, :rooms
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def price
+    digits = @price.delete"$"
+    digits.to_i
+  end
+
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -41,4 +41,12 @@ class House
   def rooms_sorted_by_area
     @rooms.sort { |room| room.area}
   end
+
+  def rooms_by_category
+    by_category = Hash.new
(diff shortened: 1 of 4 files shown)


timomitchel:
Nice job committing throughout the challenge

lib/house.rb
```diff
@@ -0,0 +1,52 @@
+class House
+  attr_reader :address, :rooms
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
```
timomitchel, line 6: Nice setup here.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def price
```
timomitchel, line 9: This is totally valid. You also could have done it in initialize on the price param. This way, this conversion happens only **once**.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+    digits = @price.delete"$"
+    digits.to_i
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    price > 500000
```
timomitchel, line 19: Simple and effective.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def rooms_from_category(category)
+    @rooms.find_all { |room| room.category == category }
```
timomitchel, line 23: The perfect enum for this task.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def area
+    @rooms.sum { |room| room.area }
```
timomitchel, line 27: Again, great job choosing the right enum.

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def test_it_can_add_rooms
+    assert_equal [], @house.rooms
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    assert_equal [@room_1, @room_2], @house.rooms
+  end
+
+  def test_it_can_tell_if_above_market_value
+    refute @house.above_market_average?
```
timomitchel, line 33: I would not use refute in this way. It will pass even if a `nil` is returned. `assert_equal false` is a better approach.

timomitchel: And, you should be testing a situation in which a `true` is returned as well.

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+    expected = {:bedroom => [@room_1, @room_2],
+                :living_room => [@room_3],
+                :basement => [@room_4]
+                }
+    assert_equal expected, @house.rooms_by_category            
+  end
```
timomitchel, line 85: All, excellent tests here.

PR #36, opened 2020-05-13
AJ TRAN

ajtran303:


test/room_test.rb
```diff
@@ -8,17 +8,52 @@ def test_it_exists
...
+    assert_equal '16', room_2.width
+  end
+
   def test_it_has_a_category
-    skip
     room = Room.new(:bedroom, 10, '13')
     assert_equal :bedroom, room.category
+
+    room2 = Room.new(:living_room, 15, '12')
+    assert_equal :living_room, room2.category
```
KathleenScriver, line 28: For this and the above example, I'd be ok with you just testing the attribute for one instance.

test/room_test.rb
```diff
@@ -8,17 +8,52 @@ def test_it_exists
...
   end
 
   def test_it_can_get_area
-    skip
     room1 = Room.new(:bedroom, 10, '13')
     room2 = Room.new(:living_room, 15, '12')
     assert_equal 130, room1.area
     assert_equal 180, room2.area
```
KathleenScriver, line 35: In this example, testing two instances is preferred since we're doing a calculation and want to ensure we can handle different inputs. Just want to illustrate the different between this test and the one above.

test/room_test.rb
```diff
@@ -8,17 +8,52 @@ def test_it_exists
...
   end
+
+  def test_it_starts_not_painted
```
KathleenScriver, line 38: Great test name!

test/room_test.rb
```diff
@@ -8,17 +8,52 @@ def test_it_exists
...
+
+    room1.paint
+    assert_equal true, room1.is_painted?
+
+    room2 = Room.new(:living_room, 15, '12')
+    assert_equal false, room2.is_painted?
+
+    room2.paint
+    assert_equal true, room1.is_painted?
+  end
```
KathleenScriver, line 58: Love that you are testing the initial state, calling another method, and then testing that the first method has a different return value. Again though, no need to do this for both instances.

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    assert_instance_of House, house
+    assert_equal 400000, house.price
+    assert_equal "123 sugar lane", house.address
+  end
+
+  def test_it_starts_with_no_rooms
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_instance_of Array, house.rooms
```
KathleenScriver, line 20: Great job! Love that you're thinking about testing data types as well. I also really appreciate the name of this test. It is very descriptive.

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    assert_empty house.rooms
+  end
+
+  def test_it_starts_with_a_details_hash
+    house = House.new("$400000", "123 sugar lane")
+
+    expected_hash = {"price" => 400000, "address" => "123 sugar lane"}
+
+    assert_equal expected_hash, house.details
```
KathleenScriver, line 29: Good job figuring out how to test a hash with minitest.

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    house.add_room(room_1)
+
+    assert_equal 1, house.rooms.size
+    assert_equal [room_1], house.rooms
+
+    house.add_room(room_2)
+
+    assert_equal 2, house.rooms.size
+    assert_equal [room_1, room_2], house.rooms
```
KathleenScriver, line 47: Good job with testing here. I would be ok with you testing the initial state, then adding all the rooms, and then testing the final array, that it has both rooms, rather than checking after calling `add_room` every time.

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    house = House.new("$400000", "123 sugar lane")
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
+    assert_equal [room_4], house.rooms_from_category(:basement)
+    assert_equal [room_3], house.rooms_from_category(:living_room)
```
KathleenScriver, line 65: Good job testing for all the cases in one test. No need to test every type of room though. If we had 5,000 different types of rooms, we wouldn't need to or want to test all of them, just one or two different ones would be sufficient.

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  def test_it_is_not_above_market_average_if_exactly_500_000
+    house = House.new("$500000", "125 syrup lane")
+
+    assert_equal false, house.above_market_average?
+  end
+
+  def test_it_is_above_market_average_if_greater_than_500_000
+    house = House.new("$600000", "126 saccha lane")
+
+    assert_equal true, house.above_market_average?
```
KathleenScriver, line 83: Great job testing both the `true` and `false` cases of this method. It would be fine though to have tested all three cases in one test. Just have all three house instances and then three assertions testing each scenario. We don't want to be overly verbose in our testing.

lib/room.rb
```diff
@@ -1,12 +1,26 @@
-class Rom
+class Room
+  attr_reader :category,
+              :length,
+              :width,
+              :painted
```
KathleenScriver, line 5: Be careful with this one. The interaction pattern did not ask for a method called `painted` so we would not want this listed in the attr_reader.

lib/house.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def details
+    @details["price"] = @price
+    @details["address"] = @address
+    @details
+  end
+
+  def above_market_average?
+    @price > 500_000
```
KathleenScriver, line 29: Great job with this method. Simple is the best way and understanding the return value of this statement is key.

lib/house.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_reader :price,
+              :address,
+              :rooms
+
+
+  def initialize(price, address)
+    @price = digit(price)
+    @address = address
+    @rooms = Array.new
+    @details = Hash.new
```
KathleenScriver, line 11: Interesting choice here to store this as an instance variable. Here's why this is problematic: @details now stores information about the price and address. What if the price changes? Now you have multiple places in your code that this would need to be updated which could introduce errors. It's better to have a method that dynamic retrieves the info and can create the hash data structure on the fly so we know it's always up to date.

lib/house.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def digit(price_str)
+    price_str.delete_prefix("$").to_i
+  end
+
+  def details
+    @details["price"] = @price
+    @details["address"] = @address
+    @details
```
KathleenScriver, line 25: Refer to the comment above for this one. I would rather you created the hash in this method with all the compiled information.

lib/house.rb
```diff
@@ -0,0 +1,32 @@
...
+    @rooms = Array.new
+    @details = Hash.new
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def digit(price_str)
+    price_str.delete_prefix("$").to_i
```
KathleenScriver, line 19: Great job researching a solution to this one.

PR #37, opened 2020-05-13
Jonathan Wilson

Jonathan-M-Wilson:


test/house_test.rb
```diff
@@ -0,0 +1,81 @@
...
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+
+    assert_equal 1900, @house.area
+  end
+
+  def test_it_has_details
+
+  assert_equal {"price" => 400000, "address" => "123 sugar lane"}, @house.details
```
KathleenScriver, line 80: With minitest, it won't let you put a hash directly into the `expected` argument. You'd need to either store it in a variable or wrap the hash in parenthesis.

test/room_test.rb
```diff
@@ -9,16 +9,23 @@ def test_it_exists
...
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_it_is_painted?
+    room1 = Room.new(:bedroom, 10, '13')
+    room2 = Room.new(:living_room, 15, '12')
+
+    refute room1.is_painted?
+  end
```
KathleenScriver, line 30: I don't see any tests for the `paint` method. We'd want to test that that changes the method `is_painted?`

PR #38, opened 2020-05-13
Ryan Laleh

RyN21:


lib/house.rb
```diff
@@ -0,0 +1,57 @@
+class House
+  attr_reader :address, :rooms
+
+  def initialize (price, address)
+    @price = price
+    @address = address
+    @rooms = []
+    @rooms_by_area = {}
+    @rooms_by_category = {}
```
memcmahon, line 9: Because these are going to be methods down below, they do not need to be instance variables.  Generally we want to have as few instance variables as possible.

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def price
+    @price.delete("$").to_i
```
memcmahon, line 13: Great job with this price method!

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    @price.delete('$').to_i > 500000
```
memcmahon, line 21: Since you have the price method above, you could clean this method up like this:

```ruby
def above_market_average?
    price > 500000
end
```

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def rooms_from_category(room_type)
+    category = []
+    @rooms.each do |room|
```
memcmahon, line 26: Great use of `each` - you could also use `find_all`

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+      category << room if room_type == room.category
+    end
+    category
+  end
+
+  def area
+    @rooms.inject(0) do |sum, x|
```
memcmahon, line 33: I don't love the variable `x` - `room` would be more descriptive.

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+      sum + x.area
+    end
+  end
+
+  def details
+    details = {price, address}
+  end
```
memcmahon, line 40: This method is actually throwing a syntax error, which makes me feel like you are not running your tests......

lib/house.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def price_per_square_foot
+    area = @rooms.inject(0) do |sum, x|
+      sum + x.area
+    end
+    squarefoot = @price.delete("$").to_f.round(2) / area
+    squarefoot.round(2)
+  end
```
memcmahon, line 48: Again, try to use more descriptive variable names, and you could reuse the `price` method here as well.

lib/room.rb
```diff
@@ -1,12 +1,22 @@
-class Rom
+class Room
+  attr_reader :category, :length, :width, :paint
 
-  def initialize(category, length, width)
+  def initialize(category, length = 10, width)
```
memcmahon, line 4: There's nothing in the interaction pattern that would indicate that the length needs to have a default value.

test/test_house.rb
```diff
@@ -0,0 +1,110 @@
...
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+    assert_equal "{'price' => 400000, 'address' => '123 sugar lane'}", house.details
```
memcmahon, line 68: A few things here - the details method should return a Hash, not a String, and that hash should have keys that are symbols:

```ruby
assert_equal ({price: 400000, address: '123 sugar lane'}), house.details
```

